PR #268, opened 2013-06-25
Implement modality of a JS dialog in app shell by using a new api.

RaymondLim:
This requires another pull request on brackets repo that makes calls to the new api in dialogs.js to have it working. https://github.com/adobe/brackets/pull/4341

A new API setModal is introduced so that Brackets can tell app shell to enforce modality when a JS modal dialog is showing. Before showing the JS dialog, we need to call this api with ``true`` argument; and after dismissing the dialog, we need to call it with ``false`` argument.

When the modality is on, all menu items will be disabled and users will not be able to click on any menu item or using its shortcuts. One exception to this on Mac is that some of the Edit menu items (Undo, Redo, Cut, Copy, Paste and Select All) will still be enabled so that users can use them in text fields in any JS dialog.

Modality is not only enforced on menus and shortcuts, but also on other UIs and shortcuts available on the window title bar. That is. X button and Alt+F4 shortcut will be disabled during the display of a JS dialog.

Although menus and commands won't be available in a JS dialog, dialog-specific shortcuts or key events will be available. For example, Cmd-D, Cmd-. and Escape key to dismiss the dialog.

### DIFF
--- a/appshell/appshell_extensions.cpp
+++ b/appshell/appshell_extensions.cpp
@@ -626,6 +626,22 @@ public:
                 responseArgs->SetString(2, parentId);
                 responseArgs->SetInt(3, index);
             }
+        } else if (message_name == "SetModal") {
+            // Parameters:
+            //  0: int32 - callback id
+            //  1: bool - on or off
+            if (argList->GetSize() != 2 ||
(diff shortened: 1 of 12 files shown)


RaymondLim:
Closing for now.

PR #333, opened 2013-09-20
Implement Brackets Portable Build on Mac and Windows

bchintx:
Allows Brackets to be run (eg. from removable media, such as a thumb drive) on any Mac or Windows computer without having to first be installed on that computer.  Implements ["Portable build" user story](https://trello.com/c/sHyY0KlB/1003-portable-build).

To set up a portable build onto a removable media storage device, follow these steps:
1. first install Brackets onto a Mac or Windows machine
2. copy the installed binaries to a new folder on your removable media.  For Mac, copy "/Applications/Brackets Sprint XX.app".  For Windows, copy the folder "\Program Files (x86)\Brackets Sprint XX".
3. in the removable media folder, next to either the .app or .exe file, create a new, empty file, named 'makePortable'

That's it!  You can now run Brackets from that removable media on any other Mac or Windows machine.

Note: you can copy both a Mac and Windows portable build onto the same removable media.  However, please be sure to create two different folders into which you copy each platform's build.

When launched, Brackets will check for the existence of the 'makePortable' file, alongside either the .app or .exe.  If it exists, then the following will occur during program execution:
1. preferences will be read from and saved in the 'Brackets' folder on the removable media;
2. extensions will be installed in and run from the 'Brackets\extensions' folder on the removable media; and
3. (on Windows) window state and size will be saved in 'Brackets\lastWindowState.dat' on the removable media.

### DIFF
--- a/appshell/client_app.h
+++ b/appshell/client_app.h
@@ -97,7 +97,7 @@ class ClientApp : public CefApp,
                                          std::wstring& filename
 #else
                                          std::string& filename
-#endif //OS_WIN
+#endif
                                          );
   static CefString AppGetAppDirectory();
   static CefString AppGetSupportDirectory();
--- a/appshell/client_app_win.cpp
+++ b/appshell/client_app_win.cpp
@@ -168,7 +168,7 @@ bool ClientApp::IsPortableInstall()
 	    isPortableInstall = (INVALID_HANDLE_VALUE != hFile) ? ISPORTABLE : ISNOTPORTABLE;
 	    ::CloseHandle(hFile);
     }
-	return (isPortableInstall == ISPORTABLE) ? true : false;
+	return (isPortableInstall == ISPORTABLE);
 }
 
 // return the name of the portable install data file


James-Parsons:
Is anyone working on this??

le717:
> FYI- I think we could go ahead and merge this now. That would allow users to create their own portable build installations. Ideally, we'd distribute the portable build as an archive available off of http://brackets.io. However, at this point, we probably shouldn't hold up this PR for any upcoming build machine or website changes.

I agree. The schematics can be worked out after it is merged, it's not like the code here depends on  how it is packaged and released.

@bchintx Quick thought, is there a reason Linux is not supported in this PR, or is the Linux version already "portable" in the sense it requires no installation?

bchintx:
@le717 Good point.  These changes would need to be made in the Linux shell as well.  However, I don't have a Linux dev environment in which I could easily port these changes.  Without these changes, then, yes, the Linux build is just "portable" as the app is already today.  However, these changes allow the preference settings to be transportable with the install onto different machines.

avindra:
+1. FYI, per the [trello card](https://trello.com/c/sHyY0KlB/1003-portable-build), 

this workaround does the trick until an official solution is implemented (replace with your own values):

```batch
msiexec /a Brackets.Release.1.2.msi /qb TARGETDIR=C:\App\Brackets
```

digideth:
I guess we are still waiting on a official solution?

Is @avindra solution the best way?

le717:
Soooo... what is happening here? Obviously Linux support has never been added, but then again Linux is historically behind Windows and Mac (see: CEF 2127). I do not think Linux-only code is imperative to finally land this code, and as @bchintx said, the exact release schematics are not required before this is eventually landed. An official portable Brackets build would still be a very nice thing to have.

rockyallen:
Will this change allow installation without admin rights?

PR #427, opened 2014-03-12
--startup-url command line switch

glenn-murray-bse:


niedarek:
Is it possible to pass custom argument on startup like file or url?
I need to use bracket-shell custom app as extension to existing application and i need pass variable like this id=12

JeffryBooher:
@niedarek Brackets isn't a general purpose application shell and doesn't allow for custom URLs or Command Line arguments.  You might want to open an issue on the forked repo https://github.com/ascom-au/brackets-shell and see if @ascom-au will add that feature or you could fork that repo and add what you want.

ficristo:
Right now brackets-shell is only for Brackets use.
If you still have some needs better to look at https://github.com/electron/electron
After all there were some discussion to switch Brackets to it too.

PR #440, opened 2014-05-05
Adding VS 2013 compile support

ligaz:
Adds compilation support for VS 2013 tools. As some Windows functions are obsolete on newer Windows I have disabled the warnings with `#pragma`s.


nethip:
@ligaz Do you have some cycles to look into this? Updating to VS 2013 would be a very nice :smile:

ligaz:
Hey @nethip I don't think I will have time to look into this. Also VS 2015 is just around the corner so it make sense to go straight to it.

ficristo:
Now we can compile with VS2015 (and I think VS2013 too).
I opened https://github.com/adobe/brackets-shell/pull/568 to discuss the changed to the script file.

PR #447, opened 2014-05-25
add a function that gets called when the menu item state was changed

eyelash:
This will make it a lot easier to implement menu item states on Linux.
Please note that I only tested this on Linux, so someone needs to verify the windows and mac versions.

### DIFF
--- a/appshell/appshell_extensions.cpp
+++ b/appshell/appshell_extensions.cpp
@@ -606,6 +606,7 @@ public:
                 bool enabled = argList->GetBool(2);
                 bool checked = argList->GetBool(3);
                 error = NativeMenuModel::getInstance(getMenuParent(browser)).setMenuItemState(command, enabled, checked);
+                MenuItemStateChanged(browser, command);
             }
         } else if (message_name == "SetMenuTitle") {
             // Parameters:
--- a/appshell/appshell_extensions_gtk.cpp
+++ b/appshell/appshell_extensions_gtk.cpp
@@ -718,6 +718,11 @@ int32 RemoveMenuItem(CefRefPtr<CefBrowser> browser, const ExtensionString& comma
     return NO_ERROR;
 }
 
+void MenuItemStateChanged(CefRefPtr<CefBrowser> browser, const ExtensionString& commandId)
+{
+    
+}
+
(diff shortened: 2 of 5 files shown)


saurabh95:
@eyelash I think this can be closed now, since we have added a similar function in https://github.com/adobe/brackets-shell/pull/602/files#diff-e00e701b36c1939d7396559780390e0aR610 
Disabling and Enabling of Menus is working fine, but I was not able to implement Checked menu items.
@eyelash @ingorichter do you have any idea on how this can be implemented?

eyelash:
Yes, I guess this can be closed. As for checked menu items, I don't know, It's been a long time. What exactly prevents you from implementing them?

saurabh95:
Actually in GTK we can't just set the flag checked to check a menu item. We have to declare menuitem as CheckMenuItem instead of MenuItem(which we are currently using) and there is no way by which we can know whether the item is of type checked or not(I mean there is no flag passed which conveys that this item needs to be CheckMenuItem)
MenuItem is the parent class of CheckMenuItem (https://developer.gnome.org/gtk3/stable/GtkCheckMenuItem.html)

eyelash:
I just had a quick look. The simplest solution would probably be to use a [GtkCheckMenuItem](https://developer.gnome.org/gtk3/stable/GtkCheckMenuItem.html) for all menu items (except separators) but I don't know what that would look like (empty checkboxes on every menu item maybe). Otherwise you would probably need to change a few things to know at menu creation time whether the menu item can be checked or not and then either create a MenuItem or a CheckMenuItem.

saurabh95:
I already tried using  GtkCheckMenuItem for all menu items (except separators) but then when you click on a menu item, it toggles the state automatically (checked to unchecked or unchecked to checked), so we end up having checked menu item for ones which were not supposed to have check.

eyelash:
Oh, I think I remember running into this problem as well a few years ago. You could also create a custom menu item but I don't think that would work with Unity as menus are drawn by the shell and not by the app. To properly solve this I think you have to change the interface for all platforms.

saurabh95:
Yeah I think we are left with that option only, will try to change the interface and will let you know :smile:

saurabh95:
Closing this as it is already added in https://github.com/adobe/brackets-shell/pull/602/

PR #457, opened 2014-06-16
Fix Issue #7846 - "Renaming a file to the same name but a different case produces an error"

bchintx:
Fixes issues:
[#7846](https://github.com/adobe/brackets/issues/7846) - "Renaming a file to the same name but a different case produces an error"
[#8127](https://github.com/adobe/brackets/issues/8127) - "Error Renaming File"

Workaround issue with [NSFileManager -moveItemAtPath:toPath:error:] that occurs renaming with a filename change that differs in case-only (eg `test.js` and `TEST.js`).  This is supported by the Mac OSX filesystem, so it should be allowed.  To workaround this shortcoming, this change uses a temporary, intermediate filename to do the rename in two steps.


FezVrasta:
May you merge this PR? It's a very annoying bug on OSX and I have compiled the master branch applying this patch and the fix works perfectly.

ingorichter:
@bchintx any objections to get this merged?

bchintx:
@ingorichter No objections at all!  Merge away, please.  :-)

PR #464, opened 2014-09-03
[REVIEW ONLY] Enable Cookies

jasonsanjose:
Fix for https://github.com/adobe/brackets/issues/8582


utkarshut:
I am trying to get cookies from server in client app that is running on bracket but cookie only comes in response header (set-cookie). How to set it to browser cookie ??

PR #491, opened 2014-12-10
Restoring the QUIT sequence on linux to that of CefClient

nethip:
Removing the destroy() function call, which in turn calls
CefQuitMessageLoop() as the right place to call this is inside
ClientHandler::OnBeforeClose().


ficristo:

PR #495, opened 2014-12-21
Support more common key names in native menus

RaymondLim:
This fixes issue https://github.com/adobe/brackets/pull/10249.

### DIFF
--- a/appshell/appshell_extensions_mac.mm
+++ b/appshell/appshell_extensions_mac.mm
@@ -1013,12 +1013,29 @@ NSUInteger processKeyString(ExtensionString& key)
         appshell_extensions::fixupKey(key, "Opt-", "")) {
         mask |= NSAlternateKeyMask;
     }
-    //replace special keys with ones expected by keyEquivalent
-    const ExtensionString del = (ExtensionString() += NSDeleteCharacter);
-    const ExtensionString backspace = (ExtensionString() += NSBackspaceCharacter);
-    const ExtensionString tab = (ExtensionString() += NSTabCharacter);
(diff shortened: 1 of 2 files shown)


appshell/appshell_extensions_mac.mm
```diff
@@ -1013,12 +1013,29 @@ NSUInteger processKeyString(ExtensionString& key)
...
+
+    unichar pageUpChar   = NSPageUpFunctionKey;
+    unichar pageDownChar = NSPageDownFunctionKey;
+    unichar homeChar     = NSHomeFunctionKey;
+    unichar endChar      = NSEndFunctionKey;
+    unichar insertChar   = NSHelpFunctionKey;
+    
     //replace special keys with ones expected by keyEquivalent
-    const ExtensionString del = (ExtensionString() += NSDeleteCharacter);
+    const ExtensionString pageUp    = (ExtensionString() += [[NSString stringWithCharacters: &pageUpChar length: 1] UTF8String]);
```
redmunds: @rlim Why can't the new keys be defined using same pattern as existing keys? For example:

```js
    const ExtensionString pageUp    = (ExtensionString() += NSPageUpFunctionKey);
```

Seems simpler and more consistent.

redmunds:
@rlim Done with code review. Works great. Just 1 comment.

appshell/appshell_extensions_mac.mm
RaymondLim: @redmunds Because they're `unichar` and we need to convert them to UTF8.

redmunds:
Looks good. Merging.

PR #498, opened 2015-01-09
fixes initial placement, move and size issues

JeffryBooher:
FIxes the following issues:

* https://github.com/adobe/brackets/issues/10208
* https://github.com/adobe/brackets/issues/9742

### DIFF
--- a/appshell/cef_host_window.cpp
+++ b/appshell/cef_host_window.cpp
@@ -168,7 +168,20 @@ BOOL cef_host_window::HandleSize(BOOL bMinimize)
     }
     SetProp(L"WasMinimized", (HANDLE)bMinimize);
 #endif
+    NotifyWindowMovedOrResized();
+    return FALSE;
+}
 
+void cef_host_window::NotifyWindowMovedOrResized()
+{
+    if (GetBrowser() && GetBrowser()->GetHost()) {
+        GetBrowser()->GetHost()->NotifyMoveOrResizeStarted();
+    }
+}
+
+BOOL cef_host_window::HandleMoveOrMoving() 
+{
+    NotifyWindowMovedOrResized();
     return FALSE;
 }
 
@@ -196,6 +209,11 @@ LRESULT cef_host_window::WindowProc(UINT message, WPARAM wParam, LPARAM lParam)
         break;
     case WM_INITMENUPOPUP:
         if (HandleInitMenuPopup((HMENU)wParam))
+            return 0L;
+        break;
+    case WM_MOVING:
+    case WM_MOVE:
+        if (HandleMoveOrMoving())
             return 0L;
         break;
     }
--- a/appshell/cef_main_window.cpp
+++ b/appshell/cef_main_window.cpp
@@ -420,23 +420,23 @@ void cef_main_window::RestoreWindowPlacement(int showCmd)
 
         EnsureWindowRectVisibility(left, top, width, height, SW_SHOWNORMAL);
 
-        wp.rcNormalPosition.left = left;
-        wp.rcNormalPosition.top = top;
+        // presumably they would all be set to CW_USEDEFAULT
+        //  but we check for any out-of-bounds value and
+        //  bypass the restore rect as to let Windows decide
+        if (left != CW_USEDEFAULT &&
+            top != CW_USEDEFAULT && 
+            height != CW_USEDEFAULT && 
+            width != CW_USEDEFAULT) {
 
-        if (width != CW_USEDEFAULT) {
+            wp.rcNormalPosition.left = left;
+            wp.rcNormalPosition.top = top;
+
             wp.rcNormalPosition.right = wp.rcNormalPosition.left + width;
-        } else {
-            wp.rcNormalPosition.right = CW_USEDEFAULT;
-        }
-
-        if (height != CW_USEDEFAULT) {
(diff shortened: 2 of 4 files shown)


appshell/cef_host_window.cpp
```diff
@@ -168,7 +168,20 @@ BOOL cef_host_window::HandleSize(BOOL bMinimize)
...
     SetProp(L"WasMinimized", (HANDLE)bMinimize);
 #endif
+    NotifyWindowMovedOrResized();
+    return FALSE;
+}
 
+void cef_host_window::NotifyWindowMovedOrResized()
+{
+    if (GetBrowser() && GetBrowser()->GetHost()) {
+        GetBrowser()->GetHost()->NotifyMoveOrResizeStarted();
```
JeffryBooher: calling this was something that needed to be added whenever the app is moved or resized

appshell/cef_main_window.cpp
```diff
@@ -108,10 +144,14 @@ BOOL cef_main_window::Create()
     int top    = CW_USEDEFAULT;
     int width  = CW_USEDEFAULT;
     int height = CW_USEDEFAULT;
+
     int showCmd = SW_SHOW;
 
     LoadWindowRestoreRect(left, top, width, height, showCmd);
 
+    // make sure the window is visible 
+    EnsureWindowRectVisibility(left, top, width, height);
```
JeffryBooher: I tested this with multiple monitors only so please test with a single monitor.

I tested this by changing various registry values to make sure that when the restore coordinates were offscreen that brackets let Windows figure out where to place the window.

Tested Maximized windows to ensure that the restored placement was correct.  I also tested that, if the restored size was off screen, Windows figures out where to place the window when it is restored.

For reference: https://github.com/adobe/brackets-shell/pull/144

JeffryBooher: also, I tested stretching the app across to monitors to ensure that would work. The origin https://github.com/adobe/brackets-shell/pull/144 implementation would constrain it to a single monitor.

JeffryBooher:
@redmunds do you mind taking a look when you have time? Thanks!

redmunds:
Found a weird case:

1. Open Brackets in secondary monitor
2. Maximize Brackets, then shutdown
3. Restart Brackets

Results:
Brackets is opened in primary monitor. Note that I did *not* disconnect or move secondary monitor.

Then, if I Restore Brackets window size it switches back to secondary monitor!

JeffryBooher:
hmmmm...That is weird indeed. Not only that but it restores back to the secondary monitor which then maximizes back on the secondary monitor.

redmunds:
Verified that [this case](https://github.com/adobe/brackets-shell/pull/498#issuecomment-69428132) is now fixed.

I also notice that Brackets window is moved from secondary monitor if it's disconnected while Brackets is running! Sweet.

redmunds:
I noticed another case:

1. Maximize Brackets on secondary monitor and close Brackets
2. Disconnect secondary monitor and restart Brackets
3. Brackets opens maximized in primary monitor as expected
4. Now click Restore button

Results:
Brackets window become really small (390x200). Brackets is unusable as an editor, but you can still grab edges and make window bigger and everything works as expected.

Expected
Would be nice if Brackets remembered my previous Restore *size* (obviously it needs to be re-*positioned*). If that can't be remembered, then maybe default size (1000x700) can be user.

Not sure if this is worth fixing.

appshell/cef_main_window.cpp
```diff
@@ -99,6 +99,60 @@ LPCWSTR cef_main_window::GetBracketsWindowTitleText()
...
+    if (xLeft < xScreen ||
+        xTop  < yScreen ||
+        xLeft + xWidth > xScreen + cxScreen ||
+        xTop + xHeight > yScreen + cyScreen) {
+
+        // something was off-screen so reposition
+        //  to the default window placement 
+        left   = CW_USEDEFAULT;
+        top    = CW_USEDEFAULT;
+        width  = CW_USEDEFAULT;
```
redmunds: Can original width and height be maintained? i.e. only set left and top to default?

JeffryBooher: Well, this has to work for both the maximized size / placement and the restored size / placement otherwise you run into that odd case so, IMO, it's better to just let the OS decide initial placement and then let the user move it wherever they want.  

This eliminates incorrectly computing the placement or placing the window on a secondary monitor (in the case they have more than 2 monitors and disconnect one or rearrange monitors.)

JeffryBooher:
@redmunds regarding the issue you ran into with maximizing, disconnecting the monitor, restarting and clicking the restore button. In this case the Window is using CW_USEDEFAULT which lets Windows decide where to put the restored window. In some cases it's just using the minimum sizes which is 320x200. I could only reproduce this while running in the debugger so I rejiggered the logic so that we just don't set the restored window placement if it's off-screen which forces windows to just use CW_USEDEFAULT (which was passed in during initial window creation).

redmunds:
@JeffryBooher This looks good on WIndows.

Have you verified that this doesn't break Mac build? I upgraded my Mac to Yosemite and can't build Brackets at the moment.

JeffryBooher:
@redmunds this is only changing Windows project files -- it doesn't affect any cross platform code so I think we're safe on Mac.

redmunds:
Merging.

PR #500, opened 2015-02-06
Fix issue #7438 (Crash on Mac Mavericks in traffic light code)

prksingh:
Fixing trafficlightbutton crash https://github.com/adobe/brackets/issues/7438 in Brackets-shell. The crash is UTR so this is not a confirmed fix.

### DIFF
--- a/appshell/TrafficLightButton.mm
+++ b/appshell/TrafficLightButton.mm
@@ -70,15 +70,15 @@ static const int ZOOM_BUTTON_TAG = 1002;
     } else if ([self tag] == ZOOM_BUTTON_TAG){
         buttonName = @"zoom";
     }
-    active = [NSImage imageNamed:[NSString stringWithFormat:@"window-%@-active",buttonName]];
-    inactive = [NSImage imageNamed:[NSString stringWithFormat:@"window-%@-inactive",buttonName]];
-    hover = [NSImage imageNamed:[NSString stringWithFormat:@"window-%@-hover",buttonName]];
-    pressed = [NSImage imageNamed:[NSString stringWithFormat:@"window-%@-pressed",buttonName]];
+    active = [[NSImage imageNamed:[NSString stringWithFormat:@"window-%@-active",buttonName]] retain];
+    inactive = [[NSImage imageNamed:[NSString stringWithFormat:@"window-%@-inactive",buttonName]] retain];
+    hover = [[NSImage imageNamed:[NSString stringWithFormat:@"window-%@-hover",buttonName]] retain];
+    pressed = [[NSImage imageNamed:[NSString stringWithFormat:@"window-%@-pressed",buttonName]] retain];
     if (closeButton) {
-        dirtyActive = [NSImage imageNamed:[NSString stringWithFormat:@"window-%@-dirty-active",buttonName]];
-        dirtyInactive = [NSImage imageNamed:[NSString stringWithFormat:@"window-%@-dirty-inactive",buttonName]];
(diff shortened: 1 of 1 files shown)


JeffryBooher:
I looked at it and the changes look OK by me but @bobeast should have a look if he has time.  Also I need to test it and @redmunds should test it on Yosemite.

JeffryBooher:
Fixes https://github.com/adobe/brackets/issues/7438

redmunds:
This branch seems to revert to CEF 1547. Is that right?

redmunds:
I can build and run Brackets on Yosemite. I did Save As a couple times and did not see any crash.

appshell/TrafficLightButton.mm
```diff
@@ -70,21 +182,21 @@ - (void)setup {
...
-        dirtyInactive = [NSImage imageNamed:[NSString stringWithFormat:@"window-%@-dirty-inactive",buttonName]];
-        dirtyHover = [NSImage imageNamed:[NSString stringWithFormat:@"window-%@-dirty-hover",buttonName]];
-        dirtyPressed = [NSImage imageNamed:[NSString stringWithFormat:@"window-%@-dirty-pressed",buttonName]];
+        self.dirtyActive = [NSImage imageNamed:[NSString stringWithFormat:@"window-%@-dirty-active",buttonName]];
+        self.dirtyInactive = [NSImage imageNamed:[NSString stringWithFormat:@"window-%@-dirty-inactive",buttonName]];
+        self.dirtyHover = [NSImage imageNamed:[NSString stringWithFormat:@"window-%@-dirty-hover",buttonName]];
+        self.dirtyPressed = [NSImage imageNamed:[NSString stringWithFormat:@"window-%@-dirty-pressed",buttonName]];
     }
-
+
```
ingorichter: nit: please remove the trailing spaces

nethip:
@prksingh Good catch and change! But this seems to be lot of change. How about just changing to just `active = [NSImage imageNamed:[NSString stringWithFormat:@"window-%@-active",buttonName] retain];` for all the variables instead of these changes. I am afraid too many releases and retains are going to put app to an inconsistent state. And also `NSImage` returned by `[NSImage imageNeeded]` is `autorelelease`ed and so the OS is not going to release it as long as some one is holding on to it's reference, unless someone calls class `[NSAutoReleasePool release]`. Nevertheless, this is a safe change as long as the no of lines of code is put to minimum.

I think the actual culprit here is `[NSImage imageNeeded]` itself. This could be easily substituted with other ways of loading `NSImage` where we are responsible for managing memory. So that way we could just rule out any memory issues, if any.

JeffryBooher:
@prksingh also @redmunds ran into an issue where I think you just need to merge master into this branch so it can be tested.  It may have been based on an older branch because I don't see the change  to revert CEF expressly made.

prksingh:
@nethip I can remove the setters and move the retain to setup(). My only worry was in case this method gets executed multiple times we may be retaining multiple copies that could cause a leak. It's great if the `autorelease` refcounting is able to handle this. Thanks for the quick review

prksingh:
@JeffryBooher @redmunds @ingorichter Thanks for looking at this. I have updated the branch from master and removed the trailing spaces. It have also updated the changes as requested by @nethip.

Please let me know if i missed anything else. Thanks!

redmunds:
Looks good on Yosemite.

nethip:
@redmunds thanks for testing it out. Is it possible for you to give it a spin on Mavericks, if you have that setup already. If not, could anyone else give it a spin on Mavericks.

peterflynn:
Awesome to see a fix for this!  I've tweaked the title a little to be more easy to understand when looking at bug/PR lists.

PR #499, opened 2015-01-22
Fix CEF 2171 build on Linux

jasonsanjose:
Replaces #489.

Manually merge changes to `cefclient` sample app  into our linux implementation.
- GYP file changes https://code.google.com/p/chromiumembedded/source/browse/branches/2171/cef3/cef.gyp
- Client app changes https://code.google.com/p/chromiumembedded/source/browse/branches/2171/cef3/tests/cefclient/

PR #501, opened 2015-02-09
fix libgcrypt dependency on debian 8 jessie

jasonsanjose:
See https://github.com/adobe/brackets/issues/10255

### DIFF
--- a/installer/linux/debian/control
+++ b/installer/linux/debian/control
@@ -5,7 +5,7 @@ Priority: optional
 Architecture: <%= arch %>
 Installed-Size: <%= size %>
 Pre-Depends: dpkg (>= 1.14.0)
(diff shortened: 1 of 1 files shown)


peterflynn:
@jasonsanjose Is this change only correct after Linux has switched to the new CEF?  (PR #499 hasn't been merged yet...)

jasonsanjose:
No, this shouldn't require the new CEF. However, I'm having some trouble building with CEF 1547 when merging this PR with master, so I'm unable to confirm that at the moment.

Romane-T:
Good morning

See also thread https://github.com/adobe/brackets/issues/10255 re: libcrypt version breaks Brackets

As per posting in that thread from Jason, installed and tested build at https://github.com/adobe/brackets/releases/download/linux-cef-2171/Brackets.Release.1.2.64-bit.deb.

Previous version of Brackets purged. libgcrypt11 purged.

Installed Brackets from above link. No errors, no warnings. Runs (so far) with no problems.

Debian Jessie with KDE as desktop environment.

With greetings

Romane

nethip:
@jasonsanjose Do you have some cycles to validate this PR? Is this good to be merged?

jasonsanjose:
Sorry, I don't have any spare cycles.

ingorichter:
Let me get my vm image back to work, then I can check this out.

brbsix:
I saw this discussed in https://github.com/adobe/brackets/issues/10255 and just want to confirm that this will not work. The *.deb* will install of course, but Brackets won't start. Even string-replacing all references to libgcrypt.so.11 with libgcrypt.so.20 in *lib/libcef.so* will not work. The current CEF has an unavoidable hard dependency on the libgcrypt11 shared library.

nethip:
@brbsix Thanks for taking some time in testing and confirming 👍 

@ingorichter Thanks for your time too! We could definently use your expertise in this issue.

jellemdekker:
Fixed by: [adobe/brackets-shell#619](https://github.com/adobe/brackets-shell/pull/619). This pull request is now obsolete.

PR #502, opened 2015-02-10
Hi dpi support in windows.

abose:
1. Adding HiDPI assets[Minimize, maximize, close, restore button] for windows at 1.25, 1.5,2 and 2.5x scale factors.
2.  Changing Application manifest to enable HiDPI support.
3.  Loading title bar icons based on current system scale factor.


nethip:
@JeffryBooher @redmunds Could you please take a look at this PR.

appshell/cef_dark_window.cpp
```diff
@@ -190,42 +190,68 @@ void cef_dark_window::InitMenuFont()
...
-    }
-    if (mPressedSysMinimizeButton == NULL) {
-        mPressedSysMinimizeButton = ResourceImage::FromResource(MAKEINTRESOURCE(IDB_MIN_PRESSED_BUTTON));
-    }
-    if (mPressedSysMaximizeButton == NULL) {
-        mPressedSysMaximizeButton = ResourceImage::FromResource(MAKEINTRESOURCE(IDB_MAX_PRESSED_BUTTON));
+	UINT scaleFactor = GetDPIScalingX();
+
+//win HiDPI - Macro for loading button resources for scale factors start 
+#define ButtonResources(scale)\
```
nethip: Move this out and put in global scope of this file.

appshell/cef_window.cpp
```diff
@@ -347,3 +347,14 @@ BOOL cef_window::TrackNonClientMouseEvents(bool track/*=true*/)
 
     return ::TrackMouseEvent (&tme);
 }
+
+//Get the Horizontal DPI scaling factor. 
+UINT cef_window::GetDPIScalingX() const
+{
+	HDC dc = ::GetDC(mWnd);
```
nethip: Check for nullness of all the pointers in this function.

appshell/cef_window.h
```diff
@@ -74,6 +74,9 @@ static __inline int RectHeight(const RECT &rIn)
 #define DCX_USESTYLE 0x00010000
 #endif
 
+//HiDPI The default logical DPI when scaling is applied in windows. see. https://msdn.microsoft.com/en-us/library/ms701681(v=vs.85).aspx
+#define DEFAULT_WINDOWS_DPI 96
```
nethip: Can we move this to the CPP counter part. That way this variable would very localized.

appshell/cef_window.h
```diff
@@ -229,6 +232,8 @@ class cef_window
     HWND GetWindow(UINT uCmd) 
     { return ::GetWindow(mWnd, uCmd); }
 
+	UINT GetDPIScalingX() const;
```
nethip: Correct the indentation.

nethip:
Also please change the indentation from tabs to spaces.

appshell/cef_window.h
abose: used tabs all along. will correct and update.

appshell/resource.h
```diff
@@ -25,18 +25,70 @@
...
+#define IDB_MAX_HOVER_BUTTON_1X            304
+#define IDB_MAX_PRESSED_BUTTON_1X          305
+#define IDB_MIN_BUTTON_1X                  306
+#define IDB_MIN_HOVER_BUTTON_1X            307
+#define IDB_MIN_PRESSED_BUTTON_1X          308
+#define IDB_RESTORE_BUTTON_1X              309
+#define IDB_RESTORE_HOVER_BUTTON_1X        310
+#define IDB_RESTORE_PRESSED_BUTTON_1X      311
+
+#define IDB_CLOSE_BUTTON_125X              312
```
redmunds: I find the naming convention `125X`, `150X`, and `250X` versus `1X` and `2X` confusing. Maybe `1_25X`, `1_50X`, and `2_50X` ?

appshell/cef_window.cpp
```diff
@@ -347,3 +351,18 @@ BOOL cef_window::TrackNonClientMouseEvents(bool track/*=true*/)
...
     return ::TrackMouseEvent (&tme);
 }
+
+//Get the Horizontal DPI scaling factor. 
+UINT cef_window::GetDPIScalingX() const
+{
+    UINT scaleDPIX = 100;
+    if(mWnd)
+    {
+        HDC dc = ::GetDC(mWnd);
```
JeffryBooher: `cef_window` has a `GetDC` method. You should use that.

JeffryBooher:
Instead of having a different resource for each scale factor, couldn't we just get a really big scale of the image (say 200%) and scale it down with GDI+ to the size we need?  I guess we would need to see how good the scaling algorithm is but if the OS adds the ability to scale to more sizes then this will become difficult to manage.  It's actually already a little difficult to manage on the resource side as @redmunds points out.

abose:
I tried scaling 2.5x images , but the downscaled images looked very bad in when scaled then down to 1.2x(the close button reduced to around 5 dots!). Also scaled up images looked pixelated. We just have 4 icons in the title bar to HiDpi-fi and the i presume all the rest are svg in cef. With these being the only icons, i think we can keep these icons in these scale factors in the microsoft way untill they come up with some vector solution.

PR #458, opened 2014-06-25
Quit Fix

nathanial:
Uncommenting the GTK delete_event code appears to fix https://github.com/adobe/brackets/issues/4611

I tested on Arch Linux 64-bit, using the latest brackets-shell and brackets versions.

PR #503, opened 2015-02-10
Fix crash on quit in CEF 2171 linux

jasonsanjose:
Note: This is a merge into `jasonsanjose/cef_2171_linux`

Merge work from PR #458 from @nathanial. Update code to work with linux 2171.

@nethip can you try this and see if it fixes the bug? It seems to work better for me. It only crashed on quit once after a few dozen tries where the previous code crashed pretty consistently in my VM.

### DIFF
--- a/appshell/appshell_extensions_gtk.cpp
+++ b/appshell/appshell_extensions_gtk.cpp
@@ -535,12 +535,12 @@ void MoveFileOrDirectoryToTrash(ExtensionString filename, CefRefPtr<CefBrowser> 
 
 void CloseWindow(CefRefPtr<CefBrowser> browser)
 {
-    if (browser.get()) {
-        isReallyClosing = true;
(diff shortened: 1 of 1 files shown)


nethip:
@jasonsanjose I had already made the change which was done by @nathanial, but I also ran into similar conditions like what you ran into. After this change the crash occurrence has reduced but the crash is still there. One of the side effects I observed with this change was to do with process quit. Though the window closes after clicking on the close button, the process was not getting killed at all(I saw this in the system monitor). Let me know if you are seeing similar results. If not we should merge this change.

jasonsanjose:
@nethip Hmm, for once I have good news. For me, when the window closes (either the X or File > Quit) the processes also go away. Maybe try rebuilding with this branch?

peterflynn:
So this is needed before PR #499 can be merged?

ficristo:

PR #504, opened 2015-02-15
std::auto_ptr will leak for array, using std::vector instead.

edwin0cheng:
A little memory leak bug fix PR.


njx:
Good catch!

nethip:
Good change.

PR #506, opened 2015-02-25
Changing permission of file xdg-open to 755. xdg-open needs to be execut...

prafulVaishnav:
...able in linux for Brower launch in MultiBrowser Preview


peterflynn:
Thanks!

PR #507, opened 2015-03-06
Fix for #10686: White line along right edge of UI on high-DPI Windows

nethip:
This is a fix contributed by @fxb. This fix address the thin white line problem that appears just below the menu bar randomly.

### DIFF
--- a/appshell/cef_dark_aero_window.cpp
+++ b/appshell/cef_dark_aero_window.cpp
@@ -816,7 +816,7 @@ LRESULT cef_dark_aero_window::WindowProc(UINT message, WPARAM wParam, LPARAM lPa
         if (mReady) {
             UpdateNonClientArea();
         }
-        break;
+        return 0L;
     case WM_NCMOUSELEAVE:
         // NOTE: We want anyone else interested in this message
         //          to be notified. Otherwise the default implementation 


nethip:
CC @JeffryBooher @redmunds @peterflynn Please review this change.

appshell/cef_dark_aero_window.cpp
```diff
@@ -816,7 +816,7 @@ LRESULT cef_dark_aero_window::WindowProc(UINT message, WPARAM wParam, LPARAM lPa
         if (mReady) {
             UpdateNonClientArea();
         }
-        break;
+        return 0L;
```
JeffryBooher: Seems to work ok but the app frame is no longer drawing in the active color when activated. 
My guess is that Windows isn't sending `WM_ACTIVATEAPP` because you're preventing the `WM_ACTIVATE` message from reaching the default Window Proc but I haven't debugged this to determine if that is indeed the case.

```C
    case WM_ACTIVATEAPP:
        mIsActive = (BOOL)wParam;
        UpdateNonClientArea();
        break;
    }
    return lr;
```

I guess technically you could just keep track of the active state from here but I wonder if there are other things that are broken.

nethip:
@JeffryBooher Thanks for pointing out this and proving more info! I will look into this.

deandersson:
@nethip to be able to get the menu bar to render in the active color we can add `mIsActive = (BOOL)wParam;`to the `WM_ACTIVATE` case aswell since the rest of the function is the same.

But as @JeffryBooher pointed out what else might be broken?
MSDN has a remark about the `WM_ACTIVATE` message, that `DefWindowProc` handles keyboard focus when the window gets activated. I tried this by opening a second instance from the debug menu and when `DefWindowProc` doesnt run the second instance never gets keyboard  focus, so you can click in the second instance and activate it but when you type it goes in the first instance..

I could not replicate this problem when switching between other processes though. I tried several variations and could not really find anything that was broken when using this fix. There are probably more variations I could try but its over midnight now.. soo nighty night! :smile:

JeffryBooher:
We should try playing around with the Window Class and make the background brush black.  It's white by default. That might minimize the White Line flicker.

PR #508, opened 2015-03-10
Abose/win hi dpi hdc leak

abose:
Fix for issue.  adobe/brackets/issues/10707:
"The function below doesn't release the drawing context it uses to get the Device Capabilities which could lead to instability -- Windows will allocate memory for new ones (unless you change the CEF Window class to use a shared DC)."

Releasing dc after calculating DPI scaling.


nethip:
LGTM @JeffryBooher Could you do a quick review of this and call out anything that is conspicuously wrong?

JeffryBooher:
LGTM

PR #509, opened 2015-03-19
implement sendCommandProgress and progress callback

zaggino:
Info here: https://github.com/adobe/brackets/pull/10761


abose:
Verified, merging.

PR #512, opened 2015-04-08
Command line installation for MAC

nethip:
This PR adds the capability to install command line tools on MAC. Here is the summary.

1) Use Authorization Services API to create an authorization object in order to install shell script
2) New shell API InstallCommandLineTools() MAC only.
3) Shell script which when invoked (via symlink) checks for /Applications/Brackets.app, and if present launches that, if not gives a chance to Finder to search and launch Brackets.app


nethip:
@prksingh Please review this.

nethip:
Adding @madanbn for review.

appshell/appshell_extensions.cpp
```diff
@@ -712,6 +712,12 @@ class ProcessMessageDelegate : public ClientHandler::ProcessMessageDelegate {
 
                 browser->GetHost()->SetZoomLevel(zoomLevel);
             }
+        }
+        else if (message_name == "InstallCommandLineTools") {
+            // Parameters:
+            //  0: int32 - callback id
```
prksingh: We could add a check for the number of arguments:
 `if (argList->GetSize() != 1) {error = ERR_INVALID_PARAMS}`

appshell/appshell_extensions_mac.mm
```diff
@@ -1175,6 +1175,154 @@ int32 SetMenuTitle(CefRefPtr<CefBrowser> browser, ExtensionString command, Exten
...
+            //  1. Remove existing symlink
+            //  2. Check for existence of the directory and create one if required.
+            //  3  Create symlink at /usr/local/bin.
+            
+            // *** Removing the existing symlink ***
+            
+            args.push_back(rmArgs);
+            args.push_back(destFile);
+            
+            toolStatus = _RunToolWithAdminPrivileges(authorizationRef, rmTool, args);
```
prksingh: Is this meant to remove Brackets executable which is added by this code earlier? Why can't we reuse it?

appshell/appshell_extensions_mac.mm
```diff
@@ -1175,6 +1175,154 @@ int32 SetMenuTitle(CefRefPtr<CefBrowser> browser, ExtensionString command, Exten
...
+                    
+                    // *** Go ahead and create the symlink now ***
+                    
+                    args.clear();
+                    
+                    args.push_back(lnArgs);
+                    args.push_back(sourceFile);
+                    args.push_back(destFile);
+                    
+                    toolStatus = _RunToolWithAdminPrivileges(authorizationRef, lnTool, args);
```
prksingh: How long is `authorizationRef` valid?

appshell/appshell_extensions_mac.mm
```diff
@@ -1175,6 +1175,154 @@ int32 SetMenuTitle(CefRefPtr<CefBrowser> browser, ExtensionString command, Exten
...
+
+
+    AuthorizationRef authorizationRef = NULL;
+    
+    // AuthorizationCreate and pass NULL as the initial
+    // AuthorizationRights set so that the AuthorizationRef gets created
+    // successfully.
+    // http://developer.apple.com/qa/qa2001/qa1172.html
+    
+    authStatus = AuthorizationCreate(NULL, kAuthorizationEmptyEnvironment,
```
prksingh: Why not include this as well in the try catch block? In case an exception is thrown by OS

appshell/appshell_extensions.cpp
nethip: Will do that.

appshell/appshell_extensions_mac.mm
nethip: That is wrong. There is not guarantee that this was installed by Brackets. Also if we update the shell script in the next release, this will make sure the symlink always points to the latest shell script.

appshell/appshell_extensions_mac.mm
nethip: It is released at the end of this function.

appshell/appshell_extensions_mac.mm
nethip: I will move this inside `try` `catch`

appshell/appshell_extensions.js
```diff
@@ -829,6 +859,20 @@ if (!appshell.app) {
...
+
+    /**
+     * Install command line scripts to make Brackets launchable from command line.
+     * Right now usage is restricted to MAC only.
+     *
+     * @param {number}
+     *
+     * @return int. The remote debugging port used by the appshell.
+     */
+     native function InstallCommandLineTools();
```
prksingh: Should this be called something Brackets specific like `InstallCommandLineToolsForBrackets` or `AddBracketsToCommandLine` 

This may just be me, but I am think 'xcode' command line tools by default when I hear command line tools :) Please feel free to ignore.

appshell/appshell_extensions_mac.mm
```diff
@@ -1175,6 +1175,154 @@ int32 SetMenuTitle(CefRefPtr<CefBrowser> browser, ExtensionString command, Exten
...
+            }
+
+        } catch (...) {
+            // This is empty as the below statements will take care of
+            // releasing authorizationRef.
+            errorCode = ERR_CL_TOOLS_SERVFAILED;
+        }
+        
+        if(authorizationRef) {
+            AuthorizationFree(authorizationRef, kAuthorizationFlagDestroyRights);
```
prksingh: This call could also return an OSStatus equivalent of error. Could we return an error here, something like 'AUTH_RELEASE_FAILED'? We could bail on that error. Is there a possibility of some other extension re-using this authref

appshell/appshell_extensions_platform.h
```diff
@@ -55,15 +62,18 @@ inline void* getMenuParent(CefRefPtr<CefBrowser>browser) {
     }
     return NULL;
 }
+inline int32 InstallCommandLineTools() { return ERR_CL_TOOLS_NOTSUPPORTED; }
```
prksingh: Are we changing windows code in this PR as well?

appshell/appshell_extensions_mac.mm
nethip: I don't think any extension developer can reuse this. The required operation has succesfully been executed. I think flagging this as an error to the user is wrong as the script is already installed at `/usr/local/bin`. We could however track this internally.

appshell/appshell_extensions_platform.h
nethip: Yes. This is to make sure that when this API is called on platforms other than Mac, we are going to return an error.

nethip:
@prksingh Thanks for reviewing. I have updated the PR with your suggested changes. Let me know if they are OK. Also please review https://github.com/adobe/brackets/issues/10830.

prksingh:
Have merged the brackets PR on brackets. Merging this now.

appshell/appshell_extensions.js
```diff
@@ -103,8 +103,38 @@ if (!appshell.app) {
...
+     */
+    appshell.app.ERR_CL_TOOLS_RMFAILED      = 13;
+    
+    /**
+     * @constant Directory, into which symlink needs to be copied, creation failed
+     */
+    appshell.app.ERR_CL_TOOLS_MKDIRFAILED   = 14;
+    
+    /**
+     * @constant The required browser is not installed
```
peterflynn: Belated nit: this comment is wrong

appshell/mac/Brackets.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Give preference to launching the one in /Applications folder
```
peterflynn: It seems like the opposite would be more useful -- if the user has manually configured their PATH to point to another copy of Brackets (e.g. a dev copy), that should take precedence over the default command-line location.

appshell/appshell_extensions_mac.mm
peterflynn: Are we 100% sure it's safe to blindly blow away any file called `/usr/local/bin/Brackets`?  Is there any chance this could collide with something else of the user's that they wouldn't want deleted?

Re the "next release" scenario: we probably don't want to make users remember to manually re-run the command every release.  But we can't auto-run it unless we know whether they've run it once before.  So do we need to record some sort of marker, either in the symlink or elsewhere, so future releases can automagically detect if the command line shortcut was already installed?

appshell/appshell_extensions_mac.mm
```diff
@@ -1175,6 +1175,152 @@ int32 SetMenuTitle(CefRefPtr<CefBrowser> browser, ExtensionString command, Exten
...
+            if( toolStatus == errAuthorizationSuccess) {
+                
+                // *** Check and create if the directory is not present ***
+                
+                args.clear();
+                
+                args.push_back(mkDirArgs);
+                args.push_back(destFolder);
+                
+                toolStatus = _RunToolWithAdminPrivileges(authorizationRef, mkDirTool, args);
```
peterflynn: If `/usr/local/bin` doesn't already exist, are we sure we should automatically create it?  Will this have the correct permissions that `/usr/local/bin` should normally have?

appshell/appshell_extensions_mac.mm
```diff
@@ -1175,6 +1175,152 @@ int32 SetMenuTitle(CefRefPtr<CefBrowser> browser, ExtensionString command, Exten
...
+
+    // convert to std::vector<char *> array.
+    std::vector<char*> argv(1 + args.size(), NULL);
+
+    for (size_t i = 0; i < args.size(); ++i)
+        argv[i] = const_cast<char*>(args[i].c_str());
+    
+    // This is a deprecated API. Apple recommends a dedicated helper to
+    // fix this issue. This ideally should be replaced by SMBless API.
+    status = AuthorizationExecuteWithPrivileges(authorizationRef, tool.c_str(),
```
peterflynn: [This API was deprecated](https://developer.apple.com/library/mac/documentation/Security/Reference/authorization_ref/#//apple_ref/c/func/AuthorizationExecuteWithPrivileges) all the way back in OS X 10.7... which means it could stop working pretty soon. Would it be safer to use an officially supported technique?  [This Stackoverflow page](http://stackoverflow.com/questions/6841937/authorizationexecutewithprivileges-is-deprecated) lists several alternatives.

appshell/mac/Brackets.sh
peterflynn: In particular, this also means that the result of running "Brackets" from the command line may not be to run the copy of Brackets where the "install command line" menu item was run -- it could be a totally different copy instead.  That seems confusing.

peterflynn: I guess the `open -a Brackets "$@"` fallback could have that problem too since I think it just does a global Info.plist search across every app bundle on the machine.  Should we first try a relative path to the Brackets executable (since this script is in the bundle right next to it)?

appshell/appshell_extensions_mac.mm
nethip: We probably need to change our shell script to pick the application from where it was run.

Users don't have to run this script to install command line tools as it is just an alias, and this is the exact reason why we are choosing the alias. The alias is pointing to actual script. Considering a very common scenario where most users install this in /Application folder, when an update is available they would just replace the current Brackets.app in the /Application folder and that ways  user does not have to re-run the script. For this common scenario, there is no need to rerun the install script. 

In other cases of using multiple instances of Brackets, that is where it could by handy rerunning the script as that would make sure the alias now points to the correct app(provided we make the above change of launching the app which encloses the script)

appshell/appshell_extensions_mac.mm
nethip: This is just a stop gap solution which some other apps in the market already using. But there is no guarantee that this would ever be supported. I just upgraded my OS to 10.10.3 and thankfully it works fine there.

This is going to be a short lived one. Migration to SMBless involves much more work like creating helper tools and managing launch deamons and launchd plist. On top of this how is Brackets effected because of this change. We should take this up in the next immediate cycles.

appshell/appshell_extensions_mac.mm
nethip: Good question! I just checked and right this is what it is on my system

screen shot 2015-04-11 at 11 22 04 am

appshell/appshell_extensions_mac.mm
peterflynn: I've spun this off as adobe/brackets#10894

PR #513, opened 2015-04-12
Command line script and renaming Brackets.sh -> brackets.sh

nethip:
This change contains the following changes 
* Renames Brackets.sh to brackets.sh 
* Made some changes to script that fixes a unit testing bug
* Update the gyp script to copy brackets.sh instead of Brackets.sh

### DIFF
--- a/appshell/appshell_extensions_mac.mm
+++ b/appshell/appshell_extensions_mac.mm
@@ -1210,7 +1210,7 @@ int32 InstallCommandLineTools()
     int32    errorCode   = NO_ERROR;
 
 
-    std::string destFile   = "/usr/local/bin/Brackets";
+    std::string destFile   = "/usr/local/bin/brackets";
     std::string destFolder = "/usr/local/bin";
 
     std::string rmTool     = "/bin/rm";
@@ -1226,6 +1226,25 @@ int32 InstallCommandLineTools()
     AuthorizationRef authorizationRef = NULL;
     
     try {
+        
+        // Determine the location of the launch script.
+        // We have this file, brackets.sh, present inside resource folder.
+        
+
+        NSString* sourcePath;
+        
+        sourcePath = [[NSBundle mainBundle] pathForResource: @"brackets" ofType: @"sh"];
+        if(!sourcePath) {
+            // If we have not found this in the bundle try the hard path.
+            NSString* bundlePath = [[NSBundle mainBundle] bundlePath];
+            sourcePath = [bundlePath stringByAppendingString:@"/Contents/Resources/brackets.sh"];
+        }
+        
+        if(!sourcePath)
+            return ERR_FILE_NOT_FOUND;
+        
+        std::string sourceFile = [sourcePath UTF8String];
+        
         // AuthorizationCreate and pass NULL as the initial
         // AuthorizationRights set so that the AuthorizationRef gets created
         // successfully.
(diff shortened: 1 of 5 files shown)


appshell/appshell_extensions_mac.mm
```diff
@@ -1226,6 +1226,25 @@ int32 InstallCommandLineTools()
...
+        
+        sourcePath = [[NSBundle mainBundle] pathForResource: @"brackets" ofType: @"sh"];
+        if(!sourcePath) {
+            // If we have not found this in the bundle try the hard path.
+            NSString* bundlePath = [[NSBundle mainBundle] bundlePath];
+            sourcePath = [bundlePath stringByAppendingString:@"/Contents/Resources/brackets.sh"];
+        }
+        
+        if(!sourcePath)
+            return ERR_FILE_NOT_FOUND;
```
peterflynn: Can this ever happen?  In other words, can `stringByAppendingString` ever return null?

appshell/mac/brackets.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+TARGET_FILE=$0
+BRACKETS_APP_NAME="Brackets.app"
+
+TARGET_PATH="$(readlink "$TARGET_FILE")"
```
peterflynn: So if I'm reading this correctly, `$0` will be the symlink (`/usr/local/bin/brackets`), and `TARGET_PATH` will be the actual location of this .sh file, inside the Brackets.app bundle?  Could we add a comment here explaining that?

appshell/mac/brackets.sh
```diff
@@ -0,0 +1,38 @@
...
+    FINAL_APP_PATH="";
+fi
+
+# Give preference to launching the script's enclosing app
+if [ -x "$FINAL_APP_PATH" ]; then
+    open -a "$FINAL_APP_PATH" "$@"
+elif [ -x "/Applications/$BRACKETS_APP_NAME" ]; then
+    open -a /Applications/$BRACKETS_APP_NAME "$@"
+elif [ -x "$HOME/Applications/$BRACKETS_APP_NAME" ]; then
+    open -a $HOME/Applications/$BRACKETS_APP_NAME "$@"
```
peterflynn: Seems like this should take precedence over the global /Applications

peterflynn:
@nethip A couple small items here, but after that it should be good to merge.

appshell/appshell_extensions_mac.mm
nethip: The documentation says `stringByAppendingString` could return `null in failure scenarios. So a check here would be better for sure.

appshell/mac/brackets.sh
nethip: I will put more comments to this file.

appshell/mac/brackets.sh
nethip: Will change this.

nethip: Chances of someone running into the first condition failure are pretty less, unless someone creates a symlink to brackets.sh outside the .app(that would be without using our `install` procedure). For e.g.: After the symlink is created, if the app, from which the symlink was created, is moved or renamed, typing `brackets` would give a `-bash /usr/local/bin/brackets: no file or directory` error, as the target cannot be reached.

peterflynn: That's a good point... and Launch Services is probably smart enough to get the user-Applications vs. global-Applications search order right on its own too, so probably all we really need are the first case and the last case.  But the code as-is now seems fine too.

peterflynn:
Looks good now -- merging

PR #514, opened 2015-04-14
Windows Installer changes for optional PATH and context menu updating along with command line script addition.

nethip:
This change involves the following

1) Made the update path and context menu addition to file and folder context menus optional.
2) The PATH varible will now be [INSTALLDIR]/command.
3) Changed project setup scripts to create commands/brackets.bat by copying it from scripts/brackets.bat
4) Changed the Gruntfile to copy the new folder command to Brackets staging area.

### DIFF
--- a/installer/win/Brackets.wxs
+++ b/installer/win/Brackets.wxs
@@ -62,14 +62,6 @@
                                Name="!(loc.ProductName) FileExt" Type="string" />
             <?endforeach?>
             
-            <!-- Add Open With Brackets to explorer's file context menu -->
-            <RegistryValue Root="HKCR" Key="*\shell\!(loc.OpenWithText)\command" Value="[INSTALLDIR]$(var.ExeName).exe &quot;%1&quot;"
-                               Type="string" />
-            
-            <!-- Add Open With Brackets to explorer's folder context menu -->
-            <RegistryValue Root="HKCR" Key="Folder\shell\!(loc.OpenWithText)\command" Value="[INSTALLDIR]$(var.ExeName).exe &quot;%1&quot;"
-                               Type="string" />
-            
             <!-- create ProgId entry -->
             <RegistryValue Root="HKLM" Key="SOFTWARE\Classes\!(loc.ProductName) FileExt\shell\open\command"
                            Value="&quot;[INSTALLDIR]$(var.ExeName).exe&quot; &quot;%1&quot;" Type="string" />
@@ -77,6 +69,22 @@
                            Value="[INSTALLDIR]$(var.ExeName).exe,0" Type="string" />
             <RegistryValue Root="HKLM" Key="SOFTWARE\Classes\!(loc.ProductName) FileExt\shell\open"
                            Name="FriendlyAppName" Value="!(loc.ProductName)" Type="string" />
+
+            <!-- Add an entry to make sure it is runnable via Run Command  -->
+            <RegistryValue Root="HKLM" Key="SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\Brackets.exe" Value="[INSTALLDIR]$(var.ExeName).exe"
+                               Type="string" />
+        </Component>
+
+        <Component Id="AddContextMenu"
+                         Guid="5E95F75D-91C4-44B6-B8A3-83005388B4BF" Directory="INSTALLDIR" KeyPath="yes">
+
+            <!-- Add Open With Brackets to explorer's file context menu -->
+            <RegistryValue Root="HKCR" Key="*\shell\!(loc.OpenWithText)\command" Value="[INSTALLDIR]$(var.ExeName).exe &quot;%1&quot;"
+                               Type="string" />
+
+            <!-- Add Open With Brackets to explorer's folder context menu -->
+            <RegistryValue Root="HKCR" Key="Folder\shell\!(loc.OpenWithText)\command" Value="[INSTALLDIR]$(var.ExeName).exe &quot;%1&quot;"
+                               Type="string" />
         </Component>
         
         <!-- Start Menu Shortcuts-->
@@ -114,7 +122,7 @@
                                  Name="PATH"
                                  Permanent="no"
                                  System="yes"
-                                 Value="[INSTALLDIR]" />
+                                 Value="[INSTALLDIR]command" />
                   </Component>
                 </Directory>
             </Directory>
@@ -131,8 +139,15 @@
             <ComponentRef  Id='StartMenuShortcut' />
 
             <ComponentRef  Id='FileAssociations' />
+        </Feature>
+
+        <Feature Id="UpdatePath" Title="Updating Path Variable" Level="1">
             <ComponentRef  Id="UpdatePath" />
-        </Feature>                  
+        </Feature>
+        
+        <Feature Id="AddContextMenu" Title="Adding Context Menus" Level="1">
+            <ComponentRef  Id="AddContextMenu" />
+        </Feature>
 
         <SetProperty Id="ARPINSTALLLOCATION" Value="[INSTALLDIR]" After="CostFinalize" />    
     </Product>
--- a/installer/win/Brackets.wxs
+++ b/installer/win/Brackets.wxs
@@ -79,11 +79,11 @@
                          Guid="5E95F75D-91C4-44B6-B8A3-83005388B4BF" Directory="INSTALLDIR" KeyPath="yes">
 
             <!-- Add Open With Brackets to explorer's file context menu -->
-            <RegistryValue Root="HKCR" Key="*\shell\!(loc.OpenWithText)\command" Value="[INSTALLDIR]$(var.ExeName).exe &quot;%1&quot;"
+            <RegistryValue Root="HKCR" Key="*\shell\!(loc.ExplorerFileOpenContextMenu)\command" Value="[INSTALLDIR]$(var.ExeName).exe &quot;%1&quot;"
                                Type="string" />
 
             <!-- Add Open With Brackets to explorer's folder context menu -->
-            <RegistryValue Root="HKCR" Key="Folder\shell\!(loc.OpenWithText)\command" Value="[INSTALLDIR]$(var.ExeName).exe &quot;%1&quot;"
+            <RegistryValue Root="HKCR" Key="Folder\shell\!(loc.ExplorerFolderOpenContextMenu)\command" Value="[INSTALLDIR]$(var.ExeName).exe &quot;%1&quot;"
                                Type="string" />
         </Component>
         
--- a/installer/win/Brackets_en-us.wxl
+++ b/installer/win/Brackets_en-us.wxl
@@ -57,5 +57,5 @@
     <String Id="ExplorerFileOpenContextMenu" Overridable="yes">Open with !(loc.ShortProductName)</String>
     <String Id="ExplorerFolderOpenContextMenu" Overridable="yes">Open as !(loc.ShortProductName) project</String>
     <String Id="AddBracketsToPath" Overridable="yes">Add &quot;!(loc.smallProductName)&quot; launcher to PATH for command line use</String>
-    <String Id="AddContextMenu" Overridable="yes">Add &quot;Open With !(loc.ShortProductName)&quot; to Explorer context menus for all files and folders</String>
+    <String Id="AddContextMenu" Overridable="yes">Add &quot;Open with !(loc.ShortProductName)&quot; to Explorer context menus for all files and folders</String>
 </WixLocalization>
--- a/installer/win/Brackets_en-us.wxl
+++ b/installer/win/Brackets_en-us.wxl
@@ -55,4 +55,6 @@
     <String Id="VerifyReadyDlgInstallText" Overridable="yes">Click Install to begin, or Cancel to exit.</String>
     <String Id="VerifyReadyDlgRemoveText" Overridable="yes">Click Remove to uninstall !(loc.ShortProductName), or Cancel to exit.</String>
     <String Id="OpenWithText" Overridable="yes">Open With !(loc.ShortProductName)</String>
+    <String Id="AddBracketsToPath" Overridable="yes">Update PATH environment variable with !(loc.ShortProductName) install location (in order to launch !(loc.ShortProductName) from command line)</String>
+    <String Id="AddContextMenu" Overridable="yes">Add &quot;Open With !(loc.ShortProductName) &quot; menu to file and folder context menus</String>
 </WixLocalization>
--- a/installer/win/Brackets_en-us.wxl
+++ b/installer/win/Brackets_en-us.wxl
@@ -19,7 +19,7 @@
     <!-- See also: product name definitions in brackets-win-install-build.xml -->
     <String Id="ProductName">Brackets</String>
     <String Id="ShortProductName">Brackets</String>
-  
+    <String Id="smallProductName">brackets</String>
     <String Id="LanguageID" Overridable="yes">1033</String>
         
     <String Id="StandardDlg_Title" Overridable="yes">[ProductName] Installer</String>
@@ -54,7 +54,8 @@
     <String Id="ExitDialogTitle" Overridable="yes">{\WixUI_Font_BannerTitle}Completed the [ProductName] Setup Wizard</String>
     <String Id="VerifyReadyDlgInstallText" Overridable="yes">Click Install to begin, or Cancel to exit.</String>
     <String Id="VerifyReadyDlgRemoveText" Overridable="yes">Click Remove to uninstall !(loc.ShortProductName), or Cancel to exit.</String>
-    <String Id="OpenWithText" Overridable="yes">Open With !(loc.ShortProductName)</String>
-    <String Id="AddBracketsToPath" Overridable="yes">Update PATH environment variable with !(loc.ShortProductName) install location (in order to launch !(loc.ShortProductName) from command line)</String>
-    <String Id="AddContextMenu" Overridable="yes">Add &quot;Open With !(loc.ShortProductName) &quot; menu to file and folder context menus</String>
+    <String Id="ExplorerFileOpenContextMenu" Overridable="yes">Open with !(loc.ShortProductName)</String>
+    <String Id="ExplorerFolderOpenContextMenu" Overridable="yes">Open as !(loc.ShortProductName) project</String>
+    <String Id="AddBracketsToPath" Overridable="yes">Add &quot;!(loc.smallProductName)&quot; launcher to PATH for command line use</String>
+    <String Id="AddContextMenu" Overridable="yes">Add &quot;Open With !(loc.ShortProductName)&quot; to Explorer context menus for all files and folders</String>
 </WixLocalization>
--- a/scripts/brackets.bat
+++ b/scripts/brackets.bat
@@ -1,0 +1,2 @@
+@echo off
+start Brackets.exe %*
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -114,7 +114,8 @@ module.exports = function (grunt) {
                             "cef_200_percent.pak",
                             "devtools_resources.pak",
                             "icudtl.dat",
-                            "libcef.dll"
+                            "libcef.dll",
+                            "command/**"
                         ],
                         "dest"      : "installer/win/staging/"
                     }
--- a/appshell.gyp
+++ b/appshell.gyp
@@ -159,6 +159,11 @@
               # Copy locales
               'destination': '<(PRODUCT_DIR)',
               'files': ['Resources/locales/'],
+            },
+            {
+              # Copy windows command line script
+              'destination': '<(PRODUCT_DIR)command',
+              'files': ['scripts/brackets.bat'],
             }
           ],
         }],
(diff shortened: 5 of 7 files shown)


nethip:
@MarcelGerber @le717 @peterflynn Please review this.

CC @ryanstewart

installer/win/Brackets_en-us.wxl
```diff
@@ -55,4 +55,6 @@
     <String Id="VerifyReadyDlgInstallText" Overridable="yes">Click Install to begin, or Cancel to exit.</String>
     <String Id="VerifyReadyDlgRemoveText" Overridable="yes">Click Remove to uninstall !(loc.ShortProductName), or Cancel to exit.</String>
     <String Id="OpenWithText" Overridable="yes">Open With !(loc.ShortProductName)</String>
+    <String Id="AddBracketsToPath" Overridable="yes">Update PATH environment variable with !(loc.ShortProductName) install location (in order to launch !(loc.ShortProductName) from command line)</String>
```
MarcelGerber: nit: space before second `&quot;`

scripts/brackets.bat
```diff
@@ -0,0 +1,2 @@
+@echo off
+start Brackets.exe %*
```
peterflynn: What does the `start ` prefix do?  In the original implementation, we were just putting INSTALLDIR on the PATH, which meant users were launching Brackets.exe _directly_ when they typed `brackets` on the command line.  So I'm wondering what will be different now.

installer/win/Brackets_en-us.wxl
```diff
@@ -55,4 +55,6 @@
...
+    <String Id="AddContextMenu" Overridable="yes">Add &quot;Open With !(loc.ShortProductName) &quot; menu to file and folder context menus</String>
```
peterflynn: Suggested string changes:

> Add "brackets" launcher shortcut to PATH for command line use
> Add "Open with Brackets" to Explorer context menu for all files and folders

Two implications of this:

* Add a new variable `loc.ProductCommandName` for the lowercase "brackets" string
* Change the "OpenWithText" string to use lowercase "with," as seen in the 2nd string above -- that's consistent with other apps and Windows Explorer itself.

But I also think we should split "OpenWithText" into two strings for files & folders:

> Open with Brackets
> Open as Brackets project

This is easy to do and adds some clarity.

Lastly, while we're splitting it, I'd suggest we use string key names that are harder to confuse with the _regular_ "Open With" menu, which the installer also deals with.  Maybe "ExplorerMenuOpenFile" & "ExplorerMenuOpenFolder"?

installer/win/Brackets.wxs
```diff
@@ -77,6 +69,22 @@
                            Value="[INSTALLDIR]$(var.ExeName).exe,0" Type="string" />
             <RegistryValue Root="HKLM" Key="SOFTWARE\Classes\!(loc.ProductName) FileExt\shell\open"
                            Name="FriendlyAppName" Value="!(loc.ProductName)" Type="string" />
+
+            <!-- Add an entry to make sure it is runnable via Run Command  -->
+            <RegistryValue Root="HKLM" Key="SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\Brackets.exe" Value="[INSTALLDIR]$(var.ExeName).exe"
+                               Type="string" />
```
peterflynn: This is the part that will fix adobe/brackets#7482, right?

Are there any other implications to adding this registry key?

peterflynn: The info at https://msdn.microsoft.com/en-us/library/windows/desktop/ee872121%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396 seems to suggest this is a pretty safe registry addition.

nethip: Yup! This exactly fixes that problem. Also it is the Microsoft's recommended way of doing it.

scripts/brackets.bat
nethip: `start` is the command(just like open on Mac). The first argument to `start` is the program name. When this command is issued with the program name and optional arguments, it first searches for PATH variable, if it can find the program there. If not, it then goes through the registry entries present at `HKLM\SOFTWARE\Microsoft\Windows\Current Version\App Paths`. 

The change now would be that when the user types in `brackets`,  it is the command(`brackets.bat`) that is going to be executed and not the exe itself. This in turn uses `start` to launch `Brackets.exe` through the entry in the `App Paths` registry key.

installer/win/Brackets_en-us.wxl
nethip: will take care of it. I was just a placeholder string.

installer/win/Brackets_en-us.wxl
nethip: 'Add "brackets" launcher shortcut to PATH for command line use' - I think this is a little confusing as we are talking about both the shortcut as well as command line. I still feel 

*Update PATH environment variable with !(loc.ShortProductName) install location (in order to launch !(loc.ShortProductName) from command line)* is more informative and is a little better to understand.

I liked all the other string changes especially `Open as Brackets project`,

And I will make changes to the variable names as well.

scripts/brackets.bat
peterflynn: I guess what I'm wondering is, what are the differences between doing that vs. just doing `..\Brackets.exe %*` instead?

installer/win/Brackets_en-us.wxl
peterflynn: @nethip How about if we take out the word "shortcut" (since you're right -- installers also often refer to Start menu shortcuts, etc., so it could be confusing)?

> Add "brackets" launcher to PATH for command line use

This is a good deal more concise than the existing string, and it's more correct since it show the command in the canonical lower-case...

scripts/brackets.bat
nethip: This doesn't work. If  we want to get this working we need to do something similar to what we did with MAC i.e. find the absolute path and the pass the absolute path to `start`. I think the script is taking the current `pwd`, from where the script was run, and then trying to find Brackets.exe in the parent folder of the that directory.

Since we are managing all of this through our Installer it should not be a problem doing a `start Brackets.exe %*` as the chances of someone editing the registry manually, are very less.

installer/win/Brackets_en-us.wxl
nethip: Yeah this looks more concise. I will change this string.

nethip:
These are the screen shots of the installer window as well the new context menus that get installed.

**New Installer Screen**

newinstallerwindow

**Context Menus**

*File Context Menu*
filecontextmenu

*Folder Context Menu*
foldercontextmenu

nethip:
Also you could try installing from the link below.
https://www.dropbox.com/s/28zok1cjz0uxkkn/Brackets%20Release%201.3.zip?dl=0

le717:
Thanks for the screenshots and test installer. I'll try to look at this for you today, @nethip.

nethip:
@le717 Sure thanks!

nethip:
@peterflynn If these changes are OK could we go ahead with the merge.
@le717 Did you get a chance to play around with the build. Let me know in case of any issues.

peterflynn:
@nethip The UI here looks great.  However, when I installed this on my Win 8 machine it did not appear to modify the PATH, and typing `brackets` in a command prompt doesn't work.  Is a reboot supposed to be needed or anything?  Fwiw, I unchecked Explorer menu, but left PATH checked.

peterflynn:
Actually, if I open an _admin_ command prompt it works.  Command prompts without admin privileges (i.e. the most common kind) have a different PATH though, without Brackets included.  That seems like a blocker for landing this :-/

nethip:
@peterflynn I just checked. In the later case the Path is getting picked up from System variables PATH. In your first case PATH got picked from user variables PATH. Does it make sense to add PATH to both system and user variables?

peterflynn:
It absolutely needs to work in the normal command prompt (people almost never run a UAC-elevated command prompt), so I'd say yes.  Either that or _only_ add to the user one and skip the system one altogether?

nethip:
@peterflynn Actually the change is very simple. In `Brackets.wxs` we just need to change this `System="yes"` to `System="no"`. This is how it will now look. Let me quickly check this and get back to you. Thanks for catching this!

```
<Environment Id="PATH"
                               Action="set"
                                Part="last"
                                Name="PATH"
                                Permanent="no"
                                System="no"
                                Value="[INSTALLDIR]command" />
```

nethip:
@peterflynn I think a reboot is still required. I changes both these settings and on some machines the PATH is not getting updated until rebooted. Can you confirm this. I checked this on some systems.

nethip:
@peterflynn @ryanstewart I figured out what the problem was. There is a problem with MSI built with WIX to not propagate the PATH environment variable change to the explorer. In order to solve this problem one has to reboot the system. This has been a classic problem and there are lot of threads on stack overflow to overcome this problem. Everyone suggests custom actions. 

So in order to solve this problem I have bootstrapped a C++ module to MSI, which gets called after the installation is finished. This C++ routine basically broadcasts a WM_STATE_CHANGE message, so that the explorer knows it needs to refresh its environment. I have created a PR for this.

https://github.com/adobe/brackets-shell/pull/515. I have raised this to be merged with this PR rather than master.

In this PR you will find a DLL that contains the C++ routine. I built this DLL from my local system and having this part of the `grunt` job is a lot of effort. So for now we could ship this as DLL in our repo. I will add `grunt` tasks to build this DLL in next release/may be just post the project file at another repo location and add instructions to build and continue with the installation process.

The following is the C++ code. It would be great if someone can do a quick review of this.

```

UINT __stdcall RefreshEnvironmentVariables(MSIHANDLE hInstall)
{
	HRESULT hr = S_OK;
	UINT er = ERROR_SUCCESS;

	hr = WcaInitialize(hInstall, "RefreshEnvironmentVariables");
	ExitOnFailure(hr, "Failed to initialize");

	WcaLog(LOGMSG_STANDARD, "Initialized.");

	// Send out the Settings Changed message - Once using ANSII...
	SendMessageTimeout(HWND_BROADCAST, WM_SETTINGCHANGE, 0, (LPARAM)"Environment", SMTO_ABORTIFHUNG, 5000, NULL);

	// ...and once using UniCode (because Windows 8 likes it that way).
	SendMessageTimeout(HWND_BROADCAST, WM_SETTINGCHANGE, 0, (LPARAM)L"Environment", SMTO_ABORTIFHUNG, 5000, NULL);

LExit:
	er = SUCCEEDED(hr) ? ERROR_SUCCESS : ERROR_INSTALL_FAILURE;
	return WcaFinalize(er);
}

```

If any one is interested, here is the link from which I took this code.
http://www.kajabity.com/2013/09/an-update-to-the-apache-maven-installers-broadcasting-wm_settingchange-to-update-environment-variables/

nethip:
Also I have posted this reposted this build again at the following location.

https://www.dropbox.com/s/28zok1cjz0uxkkn/Brackets%20Release%201.3.zip?dl=0

@peterFlynn @le717: Could you please install the build from the new location and let me know if you are running into any issues.

peterflynn:
@nethip Hmm, I'm confused -- in the previous build, the command prompt _did_ reflect the variable right away (without reboot) but only when using the admin PATH.  If we change to setting the non-admin path instead, why would that behave any differently?

Note there's a difference between:

1. Existing command prompts that are already open (before running installer) update to reflect the new PATH (without needing to reboot or close/reopen the command prompt)
2. Command prompts you open _after_ running the installer reflect the new PATH (without needing to reboot)

It seems like we already had (2) working.  It's very unlikely users expect (1) to work, since it basically never works like that when anything else updates the PATH.

[This answer](http://serverfault.com/a/8856) seems to agree:

> Any programs spawned via Explorer after this should get the updated environment, although already-running programs will not, unless they handle the setting change message.

In other words, we shouldn't have to do anything special to (2) to work -- the extra `WM_SETTINGCHANGE` is only needed if we want (1) to work, which I think is overkill since no one else does that.

le717:
I'm downloading and installing the new installer right now. TBH, I'm still not totally sold on adding Brackets the the PATH, for as we've said, literally no other Windows IDE does this.

le717:
I just took at look at this. I'll edit this message with my response, so check back in a short bit. ;)

le717:
> Existing command prompts that are already open (before running installer) update to reflect the new PATH (without needing to reboot or close/reopen the command prompt). It's very unlikely users expect (1) to work, since it basically never works like that when anything else updates the PATH.

@peterflynn is correct, any command prompts currently open are _not_ expected to get the updated environment. It's not worth extra code to do something that's not expected to happen in the first place.

installer/win/Brackets_en-us.wxl
```diff
@@ -54,5 +54,8 @@
     <String Id="ExitDialogTitle" Overridable="yes">{\WixUI_Font_BannerTitle}Completed the [ProductName] Setup Wizard</String>
     <String Id="VerifyReadyDlgInstallText" Overridable="yes">Click Install to begin, or Cancel to exit.</String>
     <String Id="VerifyReadyDlgRemoveText" Overridable="yes">Click Remove to uninstall !(loc.ShortProductName), or Cancel to exit.</String>
-    <String Id="OpenWithText" Overridable="yes">Open With !(loc.ShortProductName)</String>
-</WixLocalization>
\ No newline at end of file
+    <String Id="ExplorerFileOpenContextMenu" Overridable="yes">Open with !(loc.ShortProductName)</String>
+    <String Id="ExplorerFolderOpenContextMenu" Overridable="yes">Open as !(loc.ShortProductName) project</String>
+    <String Id="AddBracketsToPath" Overridable="yes">Add &quot;!(loc.smallProductName)&quot; launcher to PATH for command line use</String>
+    <String Id="AddContextMenu" Overridable="yes">Add &quot;Open with !(loc.ShortProductName)&quot; to Explorer context menus for all files and folders</String>
```
le717: How about "Add "Open with !(loc.ProductName)&quot; to Explorer file and folder context menus"?

installer/win/Brackets_en-us.wxl
```diff
@@ -54,5 +54,8 @@
     <String Id="ExitDialogTitle" Overridable="yes">{\WixUI_Font_BannerTitle}Completed the [ProductName] Setup Wizard</String>
     <String Id="VerifyReadyDlgInstallText" Overridable="yes">Click Install to begin, or Cancel to exit.</String>
     <String Id="VerifyReadyDlgRemoveText" Overridable="yes">Click Remove to uninstall !(loc.ShortProductName), or Cancel to exit.</String>
-    <String Id="OpenWithText" Overridable="yes">Open With !(loc.ShortProductName)</String>
-</WixLocalization>
\ No newline at end of file
+    <String Id="ExplorerFileOpenContextMenu" Overridable="yes">Open with !(loc.ShortProductName)</String>
+    <String Id="ExplorerFolderOpenContextMenu" Overridable="yes">Open as !(loc.ShortProductName) project</String>
+    <String Id="AddBracketsToPath" Overridable="yes">Add &quot;!(loc.smallProductName)&quot; launcher to PATH for command line use</String>
```
le717: The user does not need to know the exact case of the launcher (Windows is case-insensitive, after all) nor that we use a launcher at all. This makes it feel a bit technical. How about "Add !(loc.smallProductName) to the system PATH for command line launching"?

le717:
@peterflynn Just ran the installer on Win8.1 Pro x64, couldn't repo your bug. :\

nethip:
@peterflynn @le717 I tested this with multiple systems(tested on 3 systems all running Win 7 64 bit) and I am able to repro this issue consistently. I have not tried this on Win 8 though. It does not matter whether command prompt is open or not or if a new `cmd` is opened after the installation. The only way some one can solve this problem is to restart the system. So the answer @peterflynn is referring is by changing the env variables manually. Programs, like `msiexec`, need to broadcast `WM_SETTINGSCHANGE` in order for the explorer the refresh the environment, which in our WIX case is not getting done for some strange reason.

@peterflynn I ran into exactly your case where immediately after I install I was able to use `brackets` from command prompt on my system. But I then figured out that it could be something with the `env`caching. I was able to repro this after I uninstall Brackets, restart and install brackets again.

Here are the right steps to repro this.

* Uninstall `Brackets`.
* Restart the system
* Install my build that I shared earlier. (not the one I reposted with fix )
* open `command prompt` and type in `Brackets`.

After performing these steps. I will double check with a Win 8 system in the office and update you guys on that.

If we don't want to add this custom action, then we need to atleast convey the user with something like

* Add `brackets` launcher to PATH for command line use (**reboot required**)

nethip:
@le717 About the editors which add themselves to `PATH`, there are a couple of them, which I see on my system itself. I can see that there are paths to Visual Studio, Perforce appended to PATH variable. 

We now have minimized side effects by just adding the PATH to `[INSTALLDIR]\command` folder rather than to Brackets `[INSTALLDIR]` itself.

I still think the strings in the current form are concise and are easily understandable. Anything more verbose would be an overkill. In fact we had something similar to what you had proposed earlier before changing them to the current form.

ryanstewart:
@peterflynn @nethip If we need to land this to get the command prompts working, let's do it. It sounds like changing that single flag wasn't enough. 

@le717 I propose we go ahead and do the prerelease build with the checkbox checked by default and collect feedback over the week. I'll also send out a message to the Brackets-Dev list to get feedback. If we decide to roll it back and un-check it, that sounds like an easy fix.

@nethip If we un-check it, the only way a user would be able to get the command line tool to work would be to (a) add it to their path manually or (b) reinstall Brackets, right?

nethip:
@ryanstewart Yup! The only way this works is by reinstalling. We could provide ways to do this in our command line wiki that if a user wants to do this manually.

peterflynn:
@ryanstewart The main risk of adding custom native code into the installer is that _any_ bug in this code is effectively 100% bricking Brackets for users who hit it, because they've completely unable to even finish getting it installed.  Wix/MSI may be ugly and annoying, but it has been very reliable so far.

Is @nethip the _only_ person who can repro the issue where a restart is required?  It sounds like both me & @le717 are able to see the PATH update without a restart (at least in some command prompts -- see discussion above about admin vs. non-admin).  Maybe we should cast a wider net and see how widely reproducible the bug is first?  Or try to get a sense of what triggers people to hit it...

peterflynn:
@le717 I'm confused about this:

> No, non-admin prompts can access the system PATH (open a non-admin prompt and run echo %PATH%). The only different between the two paths are system wide (all user accounts) vs. current user account only. If I had to vote for which one to use, I'd say system unless there's a reason to restrict it to current user

Both command prompts have a PATH variable, yes -- but they are two different values.  On my Win 8 machine, `echo %PATH%` does not print the same value.  The admin command prompt seems to reflect the system PATH while the regular command prompt seems to reflect the user path.  Between the two, the user command prompt is the 99% use case, so we absolutely have to get that one right.

But you said you can't repro the bug -- do you mean both regular _and_ admin command prompts show Brackets in the path after installing?  How are you opening them?  (I tried both Win-R to run `cmd`, and Win-X > Command Prompt).  Are you sure you were using the installer before @nethip changed the code to set user path instead of system?

peterflynn:
Fwiw btw, I think it's fine to leave the checkboxes checked by default.

That seems to be what other apps do, and most Windows power users are pretty well trained to inspect every checkbox in an installer to make sure it's something they really want (due to the proliferation of optional adware/bloatware in installers).

ryanstewart:
@peterflynn @nethip I was able to reproduce the behavior that Prashanth saw on my Windows 8 machine. After installing without his patch I wasn't able to use the `brackets` command.

nethip:
@ryanstewart Thanks for testing this our Ryan!

@peterflynn @le717  I have not changed the `System="yes"` to `System="no"`, because it does not seem to effect anything. 

The code that is going to be executed is pretty small. In fact the method above,`RefreshEnvironmentVariabels()`, was generated by Windows Installer XML wizard in VS 2010. The only two lines of code I have added are these.

```
    // Send out the Settings Changed message - Once using ANSII...
    SendMessageTimeout(HWND_BROADCAST, WM_SETTINGCHANGE, 0, (LPARAM)"Environment", SMTO_ABORTIFHUNG, 5000, NULL);

    // ...and once using UniCode (because Windows 8 likes it that way).
    SendMessageTimeout(HWND_BROADCAST, WM_SETTINGCHANGE, 0, (LPARAM)L"Environment", SMTO_ABORTIFHUNG, 5000, NULL);

```

These are pretty safe and also the `SMTO_ABORTIFHUNG` takes care of any waiting issues. `Custom Actions` are something provided by Wix framework itself. And considering this is provided by Wix framework, it should be safe to use them. There are lot of Wix developers, who rely a lot on custom actions.
 
Now that @ryanstewart had hit this too, looks like this is more of a common scenario and needs to be fixed.

nethip:
@peterflynn Based on your comments, I have now reduced the failure conditions by always flagging a success in the custom code. Here is the new code.

```
UINT __stdcall BroadcastEnvironmentVariablesChange(MSIHANDLE hInstall)
{
	HRESULT hr = S_OK;
	UINT er = ERROR_SUCCESS;

	hr = WcaInitialize(hInstall, "BroadcastEnvironmentVariablesChange");
	ExitOnFailure(hr, "Failed to initialize");

	WcaLog(LOGMSG_STANDARD, "Initialized.");

	// Send out the Settings Changed message - Once using ANSII...
	SendMessageTimeout(HWND_BROADCAST, WM_SETTINGCHANGE, 0, (LPARAM)"Environment", SMTO_ABORTIFHUNG, 5000, NULL);

	// ...and once using UniCode (because Windows 8 likes it that way).
	SendMessageTimeout(HWND_BROADCAST, WM_SETTINGCHANGE, 0, (LPARAM)L"Environment", SMTO_ABORTIFHUNG, 5000, NULL);

LExit:
	
	er = SUCCEEDED(hr) ? ERROR_SUCCESS : ERROR_INSTALL_FAILURE;
	if (er == ERROR_INSTALL_FAILURE)
		WcaLog(LOGMSG_STANDARD, "Custom action to broadcast environment variables change has failed!");
	
	// Since this custom action is not application critical custom action, we will ignore any failures.
	return WcaFinalize(ERROR_SUCCESS);

}
```
I would like to see if we are running into problems by getting this tested by larger audience. As we are doing a prerelease build today, I think we should use this as an opportunity to see if we are running into more problems, with this new change.

I could then work with @peterflynn and @le717 for further changes, that would make it to the final release.

nethip:
I am closing this PR so that we can re-target it to release.

PR #520, opened 2015-04-21
Fixes for folder context menu appearing for folders like Recycle bin and brackets launcher script from shell emulators.

nethip:
1) Moved the registry entries in the installer to HKCR\directory from HKCR\folder as this was leading to "Open as Brackets project" at all the locations including "Recycle Bin"
2) Added a new script to command folder which takes care of launching Brackets from emulators like MINGW32 and CYGWIN.

### DIFF
--- a/installer/win/Brackets.wxs
+++ b/installer/win/Brackets.wxs
@@ -79,12 +79,25 @@
                          Guid="5E95F75D-91C4-44B6-B8A3-83005388B4BF" Directory="INSTALLDIR" KeyPath="yes">
 
             <!-- Add Open With Brackets to explorer's file context menu -->
-            <RegistryValue Root="HKCR" Key="*\shell\!(loc.ExplorerFileOpenContextMenu)\command" Value="[INSTALLDIR]$(var.ExeName).exe &quot;%1&quot;"
+            <RegistryValue Root="HKCR" Key="*\shell\!(loc.ProductName)" Value="!(loc.ExplorerFileOpenContextMenu)"
+                               Type="string" />
+
+            <RegistryValue Root="HKCR" Key="*\shell\!(loc.ProductName)" Name="Icon" Value="[INSTALLDIR]$(var.ExeName).exe"
+                               Type="string" />
+
+            <RegistryValue Root="HKCR" Key="*\shell\!(loc.ProductName)\command" Value="[INSTALLDIR]$(var.ExeName).exe &quot;%1&quot;"
                                Type="string" />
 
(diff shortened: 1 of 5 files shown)


nethip:
@MarcelGerber @peterflynn Could you take some time reviewing this. Thanks!

nethip:
Fixes https://github.com/adobe/brackets/issues/10956 and https://github.com/adobe/brackets/issues/10938

installer/win/Brackets.wxs
```diff
@@ -79,12 +79,25 @@
...
+
+            <RegistryValue Root="HKCR" Key="*\shell\!(loc.ProductName)" Name="Icon" Value="[INSTALLDIR]$(var.ExeName).exe"
+                               Type="string" />
+
+            <RegistryValue Root="HKCR" Key="*\shell\!(loc.ProductName)\command" Value="[INSTALLDIR]$(var.ExeName).exe &quot;%1&quot;"
                                Type="string" />
 
             <!-- Add Open With Brackets to explorer's folder context menu -->
-            <RegistryValue Root="HKCR" Key="Folder\shell\!(loc.ExplorerFolderOpenContextMenu)\command" Value="[INSTALLDIR]$(var.ExeName).exe &quot;%1&quot;"
+            <RegistryValue Root="HKCR" Key="directory\shell\!(loc.ProductName)" Value="!(loc.ExplorerFolderOpenContextMenu)"
```
rroshan1: Do we want to rename 'loc.ExplorerFolderOpenContextMenu' to 'loc.ExplorerDirectoryOpenContextMenu' ?

rroshan1:
Looks good to us.. Merging.
cc @prafulVaishnav

sprintr:
Is it possible to match its behavior with Windows command prompt? We can close command prompt window once we open brackets but on Git bash it will terminate the brackets process as well.

nethip:
@sprintr You could do that. But for that to work it is a little tricky.

**MING32 / Git Bash**
After installing Brackets, you could launch brackets from command prompt using the following command. `start` is a windows command which would spawn the process, passed to it as a parameter. The above statement would not work with older versions of Brackets (1.2 and below) as we did not have the registry entry at `HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths` before and this is changed with 1.3 installer. 

The `start` command usually searches for keys at the above location and tries to launch the same. 

`start Brackets.exe <file/folder name>`

**CYGWIN**
CYGWIN env does not recognize `start` `command so there is a workaround here. Use the following command.

`brackets.bat <file/folder name>`

PR #521, opened 2015-04-23
Changed launch script for bash shell to launch Brackets in a non blocking way

nethip:
`MING32/git bash` surprisingly understands `start` command. So changed the script to launch Brackets in non- blocking mode. However this does not work well with `CYGWIN`. So retained the earlier way of launching Brackets.exe in blocking mode in `CYGWIN`.

### DIFF
--- a/scripts/brackets
+++ b/scripts/brackets
@@ -8,10 +8,10 @@ TARGET_PATH="$(dirname "$0")"
 
 case `uname` in
     *CYGWIN*) OS_NAME="CYGWIN";;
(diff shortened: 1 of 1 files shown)


nethip:
@rroshan1 @sprintr Could you review this PR?

sprintr:
:cool: Works just as expected on Git bash

nethip:
@rroshan1 I will change the name of the OF. About the ".". this usually is the case with `CYGWIN`. This was something I found in my unit tests.

rroshan1:
Changes look good. Merging.

PR #522, opened 2015-04-24
Win(FR): Fix for truncation of the checkboxes in the installer window.

nethip:
Win(FR): Fix for truncation of the checkbox in the installer window.


rroshan1:
Changes look good. Merging.

PR #523, opened 2015-04-28
Fix compilation on Linux InstallCommandLineTools not in scope

trompa:
Small fix to compile in Linux. 
InstallCommandLineTools wasdefined inside getMenuParent


nethip:
Thanks for the fix ! @trompa

PR #531, opened 2015-08-04
Update note in readme

stevemao:
node-webkit is renamed to NW.js
also mention electron


abose:
Thanks @stevemao . Merging.

PR #524, opened 2015-05-25
Fix for #11112: Command line: "brackets" without arguments always opens a new instance of Brackets

nethip:
Brackets opens as a new instance, instead of using the existing running instance, if launched from command line without any arguments.

PR #534, opened 2015-09-09
Revert "Fix for #11112: Command line: "brackets" without arguments al…

nethip:
Reverts adobe/brackets-shell#524. This fixes https://github.com/adobe/brackets/issues/11672 and alike, which crept in because of this change.


abose:
Will open file with brackets open in a different instance due to this change?

nethip:
Nope! That will not get effected. What might get effected is launching Brackets from command line without arguments is now going to open new instance of Brackets.

abose:
tagging @petetnt

abose:
Looks good. Merging.

petetnt:
Reopened https://github.com/adobe/brackets/issues/11112

PR #537, opened 2015-10-09
Fix for issue #11782

nethip:
This is the fix for adobe/brackets#11782.

Apple mandates all applications to have "Enter Full Screen" menu which executes "toggleFullScreen:" and target as nil. Now if applications don't do that, they are going to add this menu to menubar.

So added the a new menu "Enter Full Screen" with shortcut as Ctrl-Cmd-F. This will make sure the OS does not go about randomly assigning shortcuts to "Enter Full Screen".

Another interesting observation is the fact that, OS automatically substitutes the language equivalent of "Enter Full Screen" for the current system language, provided we have the resource files for that languages. So in our case, this menu entry is automatically translated in cases of french and jp and we have resource files for that. Unfortunately for other languages the menu is going to be in english which is the same case with "Minimize" and "Zoom" menus. It is really weird!

This is not documented and it took a while for me to figure out what is going on.

### DIFF
--- a/appshell/mac/English.lproj/MainMenu.xib
+++ b/appshell/mac/English.lproj/MainMenu.xib
@@ -76,6 +76,12 @@
                                     <action selector="performZoom:" target="-1" id="240"/>
                                 </connections>
                             </menuItem>
+                            <menuItem title="Enter Full Screen" keyEquivalent="f" id="241">
+                                <modifierMask key="keyEquivalentModifierMask" control="YES" command="YES"/>
+                                <connections>
+                                    <action selector="toggleFullScreen:" target="-1" id="242"/>
+                                </connections>
+                            </menuItem>
                             <menuItem isSeparatorItem="YES" id="92">
                                 <modifierMask key="keyEquivalentModifierMask" command="YES"/>
                             </menuItem>
--- a/appshell/mac/French.lproj/MainMenu.xib
+++ b/appshell/mac/French.lproj/MainMenu.xib
@@ -76,6 +76,12 @@
                                     <action selector="performZoom:" target="-1" id="240"/>
                                 </connections>
                             </menuItem>
+                            <menuItem title="Enter Full Screen" keyEquivalent="f" id="241">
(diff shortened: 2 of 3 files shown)


nethip:
@ingorichter @abose @madanbn Please review this. Let us target this for 1.5

abose:
LGTM. :+1:

appshell/mac/English.lproj/MainMenu.xib
```diff
@@ -76,6 +76,12 @@
                                     <action selector="performZoom:" target="-1" id="240"/>
                                 </connections>
                             </menuItem>
+                            <menuItem title="Enter Full Screen" keyEquivalent="f" id="241">
```
madanbn: Probably a very minor one - the documentation states to add to the View menu - 

https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSWindow_Class/#//apple_ref/occ/instm/NSWindow/

If an application supports fullscreen, it should add a menu item to the View menu with toggleFullScreen: as the action, and nil as the target.

nethip: @madanbn Thanks for reviewing this! Yes you are right about this being added to the View menu. But the problem is "View" menu is not present in our resource files as it is dynamically created from the JS layer. May be we should think about how we can reorganize the View menu to comply with Apple guidelines. But for now this should work. Note that this behavior is observed only on El Capitan. I have created a sample app in Xcode on El Capitan and removed the "View" menu from the resource. The OS then then adding "Enter Full Screen" menu to "Window" menu(without any shortcut). So it looks like the OS first looks for "View" menu, if not tries to add "Enter Full Screen " to "Window" menu.

Since this PR was raised against `master`, I have raised another PR #538 targeting `release` branch. @madanbn Could you please merge that PR.

abose:
@nethip should  we close this so that the release brach can be merged to master?

nethip:
Closing this as this is now part of #538 that is targeted to release branch.

PR #538, opened 2015-10-12
Fix for issue #11782

nethip:
This is the fix for adobe/brackets#11782.

Apple mandates all applications to have "Enter Full Screen" menu which executes "toggleFullScreen:" and target as nil. Now if applications don't do that, they are going to add this menu to menubar.

So added the a new menu "Enter Full Screen" with shortcut as Ctrl-Cmd-F. This will make sure the OS does not go about randomly assigning shortcuts to "Enter Full Screen".

Another interesting observation is the fact that, OS automatically substitutes the language equivalent of "Enter Full Screen" for the current system language, provided we have the resource files for that languages. So in our case, this menu entry is automatically translated in cases of french and jp and we have resource files for that. Unfortunately for other languages the menu is going to be in english which is the same case with "Minimize" and "Zoom" menus. It is really weird!

This is not documented and it took a while for me to figure out what is going on.

### DIFF
--- a/appshell/mac/English.lproj/MainMenu.xib
+++ b/appshell/mac/English.lproj/MainMenu.xib
@@ -76,6 +76,12 @@
                                     <action selector="performZoom:" target="-1" id="240"/>
                                 </connections>
                             </menuItem>
+                            <menuItem title="Enter Full Screen" keyEquivalent="f" id="241">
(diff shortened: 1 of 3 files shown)


nethip:
Thanks @madanbn! Most of the conversation took place at #537. Please refer to this PR for discussion on this issue. I had to close it as that PR was targeted to `master` which will make release management harder.

PR #540, opened 2015-10-23
Updates to support CEF 3.2454

andrewleech:
I expect that this will not be an immediately desired change in master, however I needed to upgrade CEF to get access to new features and simply want to share the minor changes needed to get a build.


abose:
@nethip

nethip:
@andrewleech You might want to checkout my branch prashant/cef-upgrade-latest. Brackets-shell is migrated to 2357 on this branch. However I could not upgrade to the latest cef because of some build dependencies.  Do watch out that branch for more updates. Another thing to note here is that, starting from 2357, CEF is going to be supporting only 64 bit on MAC. This change has already been done in  my branch. Do checkout this branch and let me know if you are facing any issues.

nethip:
@andrewleech Thanks for showing your interest in contributing to Brackets. We have now upgraded CEF to 2623.

PR #541, opened 2015-12-11
Fix missing libgcrypt.so.11

brbsix:
This is a fix for https://github.com/adobe/brackets/issues/10255


abose:
Tagging @nethip

brbsix:
I was just reviewing this and realized what I had submitted was not going to work for 32-bit systems as it only included the 64-bit library (and libgcrypt is not architecture independent). I think I've addressed that here but I'm not really sure. Can someone take a look?

ficristo:
IMO embedding the libraries is not the right fix. 
Have you looked at https://github.com/adobe/brackets-shell/pull/501?
You could try and report back if that fix the issue.

brbsix:
Certainly, embedding libraries is not a desirable solution. However long story short, current releases are not compatible with libgcrypt20. The 1547 CEF has hardcoded `libgcrypt.so.11`. Search/replace of hardcoded strings in CEF binary or symlink from `libgcrypt.so.20` to `libgcrypt.so.11` has all been tried and is unsuccesful.

CEF 2171 supports libgcrypt20. Until it is ready,
options are to release software that will not run on modern distros or embed the requisite libraries. I think the latter is the obvious choice. Several other projects that use CEF (to include atom) have come to the same conclusion.

nethip:
@brbsix Thanks for this PR! We wanted to merge this PR to fix the libcryp.so.11 dependency issue. Unfortunately we could not merge this as we had a "NO GO" from our legal because of the licensing issues with packing `libcrypt.so.11` with the installer.

And you are right about fixing our installer scripts and that would be part of upgrading CEF to latest version for Linux. Unfortunately we don't have that kind of bandwidth to work on the CEF migration on Linux. 

I would be more than happy to help you if you could take up this task.

ficristo:
Thank you both to have explained the current status!

PR #542, opened 2015-12-11
Switched from JSHint to ESLint

ficristo:
Similar work to the one in Brackets.
I've added a `.bracket.json` file so I could change the prefered linter. (And remove the inline jslint declarations)
I've added an `.eslint.json` file under `tasks/` because Grunt isn't compatible with the rule `no-invalid-this`.
There should not be any change in behaviour.


appshell/node-core/DomainManager.js
```diff
@@ -21,15 +21,10 @@
...
 
-/*jslint vars: true, plusplus: true, devel: true, nomen: true, indent: 4,
-maxerr: 50, node: true */
-/*global */
-
 (function () {
     "use strict";
         
     var util              = require("util"),
-        server            = require("./Server"),
```
zaggino: why this change?

tasks/set-release.js
```diff
@@ -48,7 +47,6 @@ module.exports = function (grunt) {
             packageJSON                 = grunt.file.readJSON(packageJsonPath),
             winInstallerBuildXmlPath    = "installer/win/brackets-win-install-build.xml",
             buildInstallerScriptPath    = "installer/mac/buildInstaller.sh",
-            wxsPath                     = "installer/win/Brackets.wxs",
```
zaggino: is this a valid change?

ficristo:
@zaggino They are both unused.

zaggino:
Yep, but even requiring them can have side-effects, like parts of the module being executed, event listeners attached and so on. Are you sure not requiring them won't break anything? Instead of `server            = require("./Server")` you can do just `require("./Server")` to be sure nothing breaks.

ingorichter:
What is the status of this? Any progress? @ficristo @zaggino.

zaggino:
Seems finished to me

ficristo:
I've squashed the commit and enabled the `no-trailing-spaces` rule too.
Should be ready.

nethip:
@ficristo I tried playing with this branch. There is one side effect because of this change. After building brackets-shell with this branch and opening brackets-shell source folder, with the built shell, JS hints don't appear anymore.

screen shot 2016-06-21 at 10 54 39 pm

The below snapshot is with the shell built from the release branch.

screen shot 2016-06-21 at 10 55 24 pm

Also I have observed that if I open any file under node-core with any other project, JSHints kick in.
screen shot 2016-06-21 at 10 56 45 pm

ficristo:
I cannot replicate the issue on the node-core side.
As of now there is a newer version of ESLint which will rise some more warnings and would like to update to it.
But before can we get a consensus on what to do with these:
https://github.com/adobe/brackets/issues/11984
https://github.com/adobe/brackets/pull/11988
(For not seeing JSLint anymore: we need the ESLint extension in core or I need to reenable JSLint like I did on Brackets side)

ficristo:
I've reintroduced JSLint so we can merge this.
I removed the IIFE and so deindented the code, that's why you see a lot of changes.
You can look at this link https://github.com/adobe/brackets-shell/pull/542/files?w=1 to see code changes only.

zaggino:
Gone through it, LGTM

PR #547, opened 2016-02-01
Remove a number of unnecessary file associations

le717:
Brackets is a front-end web code editor. The Windows installer, however, registers a ton of file associations that were too generic, should belong to another IDE, or were not at all related. Combined with lack of an installation option to choose what should be registered, issues such as adobe/brackets#6073 have been created.

One fix mentioned (and agreed with multiple times) is to drop the aforementioned associations from being registered. This PR does exactly that, dropping, Python C/C#/C++, Ruby, Java (and like-Java), Perl, and generic extensions (e.g., properties, ini, txt). I left PHP in-place.

With this change, the number of associations registered drops from 92 to 57. That is still a very high number and could still be lowered, but it is already a drastic improvement.

### DIFF
--- a/installer/win/Brackets.wxs
+++ b/installer/win/Brackets.wxs
@@ -50,7 +50,7 @@
                            Value="SOFTWARE\$(var.RegistryRoot)\Capabilities" Type="string" />
 
             <!-- File associations -->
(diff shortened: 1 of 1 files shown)


ryanstewart:
Thanks @le717. I don't see any reason not to merge this from an ideal user-experience aspect. @nethip who has the most shell experience on the team to determine if this PR will cause any issues?

nethip:
Sorry for being really late on this thread. 
@ryanstewart Shouldn't be an issue. We could go with the merge.

@le717 Thanks for this PR.

PR #549, opened 2016-02-06
VS2015: fix some warnings

ficristo:
The warnings fixed were:
- GetVersionEx is deprecated and so will raise warning as errors. There were two case: in one it checked if on Windows Vista or later in the second if on Windows XP or later.
AFAICT the platforms supported are from Vista onwards: I removed the checks.
- GdiPlus.h inclusion will raise some warnings so I disabled them
- there were some redeclaration warnings

/cc @nethip

### DIFF
--- a/appshell/appshell_extensions_win.cpp
+++ b/appshell/appshell_extensions_win.cpp
@@ -422,66 +422,35 @@ int32 ShowOpenDialog(bool allowMultipleSelection,
     ConvertToNativePath(initialDirectory);
 
     if (chooseDirectory) {
-        // check current OS version
-        OSVERSIONINFO osvi;
-        memset(&osvi, 0, sizeof(OSVERSIONINFO));
-        osvi.dwOSVersionInfoSize = sizeof(OSVERSIONINFO);
-        if (GetVersionEx(&osvi) && (osvi.dwMajorVersion >= 6)) {
-            // for Vista or later, use the MSDN-preferred implementation of the Open File dialog in pick folders mode
-            IFileDialog *pfd;
-            if (SUCCEEDED(CoCreateInstance(CLSID_FileOpenDialog, NULL, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(&pfd)))) {
-                // configure the dialog to Select Folders only
-                DWORD dwOptions;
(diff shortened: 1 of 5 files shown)


petetnt:
Nice to see someone working on the VS2015 support: it's one of the two pain points of not being able to contribute so much on the `brackets-shell` side :+1: (Other one being that I don't have that much native coding experience but that's just a sidematter :joy_cat:

nethip:
@ficristo Highly appreciate this.

appshell/appshell_extensions_win.cpp
```diff
@@ -422,66 +422,35 @@ int32 ShowOpenDialog(bool allowMultipleSelection,
...
-                                // Add directory path to the result
-                                std::wstring wstrName(lpwszName);
-                                ExtensionString pathName(wstrName);
-                                ConvertToUnixPath(pathName);
-                                selectedFiles->SetString(0, pathName);
-                                ::CoTaskMemFree(lpwszName);
-                            }
-                            psi->Release();
+        IFileDialog *pfd;
+        if (SUCCEEDED(CoCreateInstance(CLSID_FileOpenDialog, NULL, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(&pfd)))) {
```
nethip: 👏

nethip: Is the following code VS 2010 and VS 2013 neutral or does it have any hard dependency on VS 2105?

ficristo: I simply deleted the old code and reindented.
I don't think there should be any dependency on VS2015.
The code was platform dependent, so if that platforms are not supported anymore should be fine both on VS2010 and VS2015 (I hope :tongue:)

nethip: Oh I did not pay close attention to the code being indented. Then I think it is just fine.

ficristo: If you add `w=1` to an url on github you'll not see the whitespaces changes.
https://github.com/adobe/brackets-shell/pull/549/files?w=1
It can be useful to times like this

nethip: @ficristo Thanks for the tip! That is sure useful.

nethip:
Thanks @ficristo

PR #553, opened 2016-05-15
[MAC] Fix for open file dialog being jerky

nethip:
This PR addresses the following items.

 - The file open/save dialog jerks a bit when launched from Brackets. The reason for that being we are not honouring the `completionHandler` semantics, when launching the dialog using `beginSheetModalForWindow`. After installing `completionHandler` block, I realised it is not straight forward, as we would have returning by then. So had to make changes which involved caching the reference of the browser, so that the same could be used inside the `completionHandler` block. Also took care of re-entrant scenarios by using a map, to cache the requests and send the response to the right corresponding browser.

 - When multiple windows like, new brackets window, or devtools, were open, the entries were missing from the *Window* menu. So made sure, we have a menu entry for all open windows under *Windows* menu on MAC.

### DIFF
--- a/appshell/appshell_extensions_mac.mm
+++ b/appshell/appshell_extensions_mac.mm
@@ -51,6 +51,27 @@ int const debugPort = 9222;
 NSString* debugPortCommandlineArguments = [NSString stringWithFormat:@"--remote-debugging-port=%d", debugPort];
 NSString* debugProfilePath = [NSString stringWithFormat:@"--user-data-dir=%s/live-dev-profile", ClientApp::AppGetSupportDirectory().ToString().c_str()];
 
+// These are required for [NSSavePanel beginSheetModalForWindow]
+// We need to store the panel to browser details
+// in a map as that could be the panel is the only
+// variable accessible inside the completionHandler
+// block. Also the map is neccessary for handling
+// reentrant scenarios. E.g.file open/save dialog could
+// be requested from from multiple Brackets windows.
+// Earlier RunModal was making sure none of the opened
+// NSWindows are clickable when running the runModal of
+// the modal sheet. With completionHandler coming into
+// picture, that is not the case anymore.
+
+struct CefBrowserDetails
+{
+    CefRefPtr<CefProcessMessage> m_response;
+    CefRefPtr<CefBrowser>        m_browser;
+};
+
+static  std::map<NSSavePanel *, CefBrowserDetails>           s_panel_to_browser;
+typedef std::map<NSSavePanel *, CefBrowserDetails>::iterator panel_iterator;
+
 ///////////////////////////////////////////////////////////////////////////////
 // LiveBrowserMgrMac
 
@@ -375,12 +396,13 @@ int32 OpenURLInDefaultBrowser(ExtensionString url)
     return NO_ERROR;
 }
 
-int32 ShowOpenDialog(bool allowMulitpleSelection,
+void ShowOpenDialog(bool allowMulitpleSelection,
                      bool chooseDirectory,
                      ExtensionString title,
                      ExtensionString initialDirectory,
                      ExtensionString fileTypes,
-                     CefRefPtr<CefListValue>& selectedFiles)
+                     CefRefPtr<CefBrowser> browser,
+                     CefRefPtr<CefProcessMessage> response)
 {
     NSArray* allowedFileTypes = nil;
     BOOL canChooseDirectories = chooseDirectory;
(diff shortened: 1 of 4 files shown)


nethip:
Tagging @abhijitApte @JeffryBooher @vdhawal for review. 

I have done extensive unit testing. So it would be good to target this for Brackets 1.7.

appshell/appshell_extensions_platform.h
```diff
@@ -97,6 +98,21 @@ int32 ShowSaveDialog(ExtensionString title,
                        ExtensionString initialDirectory,
                        ExtensionString proposedNewFilename,
                        ExtensionString& newFilePath);
+#else
+void ShowOpenDialog(bool allowMulitpleSelection,
```
abhijitapte: I'm of the opinion that we should try using unified signatures as much as possible.

appshell/client_handler_mac.mm
```diff
@@ -61,6 +61,8 @@
   if ([window styleMask] & NSFullScreenWindowMask) {
       [window setTitle:str];
   }
+    
+  [NSApp changeWindowsItem:window title:str filename:NO];
```
abhijitapte: what is this change for? Debug tools?

appshell/appshell_extensions_platform.h
nethip: Yeah I thought of that. It is just that we need to unnecessarily process arguments and return types for the sake of unification.

appshell/client_handler_mac.mm
nethip: This to make sure we have a menu entry under Windows menu. As we are now nuking the window title, the menu entry is not getting added under 'Windows' menu. So I had to explicitly call this to make sure we have a menu entry. 

Please refer to [this](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSApplication_Class/#//apple_ref/occ/instm/NSApplication/changeWindowsItem:title:filename:
) link for documentation

nethip:
@abhijitapte

appshell/appshell_extensions_mac.mm
```diff
@@ -408,23 +430,48 @@ int32 ShowOpenDialog(bool allowMulitpleSelection,
...
+            
+            CefRefPtr<CefProcessMessage> _response = _browserDetails->second.m_response;
+            CefRefPtr<CefBrowser>        _browser  = _browserDetails->second.m_browser;
+            
+            if(_response && _browser){
+
+                NSArray *urls = [openPanel URLs];
+                CefRefPtr<CefListValue> selectedFiles = CefListValue::Create();
+                if (returnCode == NSModalResponseOK){
+                    for (NSUInteger i = 0; i < [urls count]; i++) {
```
abhijitapte: Shouldn't we be using for (id url in urls) instead?

appshell/appshell_extensions_mac.mm
```diff
@@ -408,23 +430,48 @@ int32 ShowOpenDialog(bool allowMulitpleSelection,
...
+                        selectedFiles->SetString(i, [[[urls objectAtIndex:i] path] UTF8String]);
+                    }
+                }
+
+                // Set common response args (error and selectedfiles list)
+                _response->GetArgumentList()->SetInt(1, NO_ERROR);
```
abhijitapte: Shouldn't these lines be within the if (_response && _browser) condition?

appshell/appshell_extensions_mac.mm
```diff
@@ -408,23 +430,48 @@ int32 ShowOpenDialog(bool allowMulitpleSelection,
...
+    browserDetails.m_response = response;
+    
+    s_panel_to_browser[openPanel] = browserDetails;
+    
+    [openPanel beginSheetModalForWindow:[NSApp mainWindow] completionHandler: ^(NSInteger returnCode)
     {
-        NSArray* urls = [openPanel URLs];
-        for (NSUInteger i = 0; i < [urls count]; i++) {
-            selectedFiles->SetString(i, [[[urls objectAtIndex:i] path] UTF8String]);
+        panel_iterator _browserDetails;
```
abhijitapte: is it possible to use self and avoid the map altogether?

appshell/appshell_extensions_mac.mm
```diff
@@ -436,17 +483,39 @@ int32 ShowSaveDialog(ExtensionString title,
...
+    
+    s_panel_to_browser[savePanel] = browserDetails;
+    
+    [savePanel beginSheetModalForWindow:[NSApp mainWindow] completionHandler: ^(NSInteger returnCode)
+     {
+         panel_iterator _browserDetails;
+         _browserDetails = s_panel_to_browser.find(savePanel);
+         if(_browserDetails != s_panel_to_browser.end()){
+             
+             CefRefPtr<CefProcessMessage> _response = _browserDetails->second.m_response;
```
abhijitapte: when would this map entry be deleted? I don't see erase() here

appshell/appshell_extensions_mac.mm
abhijitapte: I'm not sure how the blocks work, can we capture both browser and response parameters?
We can capture variables in Lambda (in C++). Something similar available here?

abhijitapte: Luckily both are CefRefPtrs.

appshell/appshell_extensions_mac.mm
nethip: Good catch Abhijith 👍 I will make the change. I would have forgotten to remote that.

appshell/appshell_extensions_mac.mm
nethip: @abhijitapte Which lines are you referring to?

appshell/appshell_extensions_mac.mm
nethip: We could use, but this piece of code was just borrowed from older way of doing this. Do you see any advantages of using `for(id url in url;s)`?

nethip:
@abhijitapte I have now incorporated all your code review comments. Actually I was modelling everything based on [this](https://github.com/adobe/brackets-shell/blob/master/appshell/appshell_extensions_mac.mm#L684)

But I had a look at objective C `blocks` and did some experiment about accessing local enclosed scope variables. Looks like, this is possible inside a `block`. Please have a look at the changed source and let me know if you have any more comments.

appshell/appshell_extensions_mac.mm
```diff
@@ -408,23 +409,36 @@ int32 ShowOpenDialog(bool allowMulitpleSelection,
     
     [openPanel setAllowedFileTypes:allowedFileTypes];
     
-    [openPanel beginSheetModalForWindow:[NSApp mainWindow] completionHandler:nil];
-    if ([openPanel runModal] == NSOKButton)
+    // cache the browser and response variables, so that these
+    // can be accessed from within the completionHandler block.
+    CefRefPtr<CefBrowser>        _browser  = browser;
+    CefRefPtr<CefProcessMessage> _response = response;
```
abhijitapte: Can't we get rid of these two locals?

appshell/appshell_extensions_mac.mm
```diff
@@ -436,17 +450,29 @@ int32 ShowSaveDialog(ExtensionString title,
...
-    {
-        NSURL* theFile = [savePanel URL];
-        absoluteFilepath = [[theFile path] UTF8String];
 
-    }
-    [NSApp endSheet:savePanel];
+    // cache the browser and response variables, so that these
+    // can be accessed from within the completionHandler block.
+    CefRefPtr<CefBrowser>        _browser  = browser;
+    CefRefPtr<CefProcessMessage> _response = response;
```
abhijitapte: same here, can't we get rid of these two locals?

appshell/appshell_extensions_mac.mm
nethip: I am sure compiler will anyway optimize and will get rid of these two variables anyway.

appshell/appshell_extensions_mac.mm
nethip: Please refer to my comment above.

abhijitapte:
LGTM

swmitra:
LGTM. Great job again @nethip :+1: 
Thanks a lot @abhijitapte for helping us with the review process :+1: 
Merging the PR.

PR #558, opened 2016-05-31
Moved some icons under appshell/res/1x folder

ficristo:
I was a bit confused by these icons not be in a specific folder.

Side note: why only the icons with 1x resolution are mentioned in gyp?


nethip:
@ficristo I am not sure if this change ever gets baked into the product. I see that we have manually added resources, on Windows, to `cefclient.rc` file.

ficristo:
So they aren't needed to be defined on the gyp side?
I'm a bit confused how this should work...

nethip:
I don't have a Windows machine at my Home. I will check and let you know tomorrow, when I get to the office.

nethip:
@ficristo I was referring to this specific code. https://github.com/adobe/brackets-shell/pull/558/files#diff-a4a63dee4259dd6b62c8cccceb28a7daR194, inside `appshell.gyp`. I think they are there for reference purposes as the actual resources are defined inside `cefclient.rc`

About the change as such, I think the general convention is, if there is no scale factor appended to an image, it is considered 1x so I am not sure if want to move them under 1x folder.

ficristo:
I still think it is a bit confusing how it is now but I'm fine to close this PR.
It is not so important.

ficristo:
I close this stale PR.

PR #559, opened 2016-06-02
Moved client_switches files under a new common folder

ficristo:
So to match the cefclient example.


nethip:
Thanks for the PR @ficristo . I have built this on MAC and it goes fine. LGTM 👍

PR #562, opened 2016-06-25
[Review Only] Port autogen_client_side import to appshell_paths.gypi

ficristo:
The translator files give me some problems on Linux.

```
>> make: *  No rule to make target 'out/Release/obj.target/libcef_dll_wrapper/libcef_dll/ctocpp/test/translator_test_ctocpp.o',
needed by 'out/Release/obj.target/libcef_dll_wrapper.a"'.  Stop.
```

@nethip @ingorichter @jasonsanjose any idea of why?


vishwajithsandaru:
I am also getting the same error, and i look for appshell_paths.gypi, and it is much more different than the above, i am building this on opensuse, but when I try to build branch "linux-1547" it builds successfully, bt in this case i want to build the 'master', but it fails with the above error when building,

ficristo:

PR #564, opened 2016-08-10
Change path of xdg-open to match opn (for linux-1547)

ficristo:
For linux-1547 branch
See https://github.com/adobe/brackets/pull/12641


petetnt:
Added the branch name for the title for clarity.

PR #569, opened 2016-08-16
Update .gitignore

ficristo:
I generalize I bit some entries and added some other for VS and VSCode.


ingorichter:
LGTM. Thanks for updating.

PR #572, opened 2016-08-23
make sure npm dependencies are installed in the www folder

zaggino:
required for https://github.com/adobe/brackets/pull/12647

### DIFF
--- a/tasks/build.js
+++ b/tasks/build.js
@@ -127,7 +127,8 @@ module.exports = function (grunt) {
                     "git checkout " + this.data.branch,
                     "git pull origin " + this.data.branch,
                     "git submodule sync",
-                    "git submodule update --init --recursive"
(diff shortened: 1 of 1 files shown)


ficristo:
I don't think this is correct.
In the `dist` folder I don't see any `node_modules`: Brackets build system doesn't copy them.
Running `grunt build` from Brackets repo should be sufficient to see it.
What I think is a better approach is to rename `src/config.json` to `src/package.json` and then run `npm install --production` inside the `dist` folder.
(After all I feel something is not right to see a npm command inside a task named `git`)

zaggino:
This is fixed in https://github.com/adobe/brackets/pull/12647
Commit https://github.com/adobe/brackets/pull/12647/commits/4e17c9826b9f1824b057d3f111ce6092004d90e5
Closing

PR #577, opened 2016-09-14
Remove define OS_*

ficristo:
Now they are always checked in `config.h` and I made sure is always included in Brackets specific files that use those macros.


ficristo:
I've compiled this on:
Windows10 with VS2015u3
OSX 10.11.6 with XCode 7.3.1

PR #579, opened 2016-09-17
[WIP] Update to last version of cefclient

ficristo:
Here my last failed attempt to update the cefclient.
Posting here for backup.


ficristo:
With this I can compile on Linux but breakage is expected.
I was hoping to at least see brackets running in the window instead I can see only a blank page.
But looking at the developer console the scripts are there.
I tryed to load a very simple html and it loads fine.

@nethip @ingorichter do you have an idea why the html doesn't show up?

nethip:
@ficristo can you try loading another webpage and see if you are seeing this issue? Or does this happen only when you load index.html residing inside `src` folder?

ficristo:
Something like the following worked.
```html
<p style="color: red">Greetings</p>
<button onclick="alert('hello')">Click me</button>
```

While this PR (IMHO) contains some good things I had some problems on Windows.
When I tryed to comment `client_handler` and all `cef_windows` things the window didn't show up.
Maybe I'll try a different approach...

ficristo:
I started another branch: https://github.com/ficristo/brackets-shell/tree/scratch-2
There I can run all 3 platforms but it still misses things, like the dark ui.
I'll see if I can move at least Linux completely and then I wonder if we could create a new linux branch (like linux-1547 but with the new client). Still a long way for me.

PR #582, opened 2016-09-28
Use again clienthandler to show devtools

ficristo:
The current code will crash on the cefclient example code in 2623.
It seems to me the code was changed to workaround a bug in CEF, so this actually restore a previous behaviour.


ficristo:
I need to check more this one: there was an error in the console which I'm not sure if related to this PR.
Plus it will be necessary when (and if) we'll move to the new cefclient code.

ficristo:
In #589 I duplicated the `appshell_extensions.cpp` in a new file `appshell_process_message.h` where I done this change.
So this is not necessary anymore.

PR #584, opened 2016-10-02
Import cefclient-2623 code and include it on Linux build only

ficristo:
Since it is too hard for me to update all 3 platforms, now I would like to try to run only Linux with the newer cefclient code.
I removed the tests and osr files and updated the imports, otherwise the code is a copy and paste.
Actually doesn't compile without https://github.com/adobe/brackets-shell/pull/562.


ficristo:
I forgot to say, since I'm looking on at Linux, I didn't import the code specific for the other platforms.

ficristo:
I noticed some references to `test_runner` which is define in a file not included: I've removed or commented them.
To be correct, I still cannot compile on Linux, but at least it doesn't introduce more errors.
I could include all these files on other platforms but on Windows I had some problems compiling files with same name (`appshell/client_app.h` and `appshell/common/client_app.h` for example)

PR #588, opened 2016-10-06
Move more things from cefclient_gtk to appshell_helpers

ficristo:
As usual it doesn't compile but I hope to be able to compile in the near future.


ficristo:
While working on another PR I noticed some errors, ops.
Rebased.

PR #591, opened 2016-10-23
Support AppGetChromeVersionString and AppGetProductVersionString on Linux

ficristo:
This is the updated version of https://github.com/adobe/brackets-shell/pull/511 
Rebasing that was a bit of a pain so I created it anew.


ficristo:
Rebased on top of master and to point to 1.9.0.0 version

PR #602, opened 2017-04-20
Added support of Native Menus in Linux

saurabh95:
To test this `global.brackets.nativeMenus`  in https://github.com/adobe/brackets/blob/master/src/utils/Global.js#L84 needs to be set to true.
// cc @abhijitapte for review.

@eyelash Could you also review the part added by me?

Fixes https://github.com/adobe/brackets/issues/7236

Screenshot for Ubuntu 16.04

screenshot from 2017-04-23 13-04-15


eyelash:
it's been 3 years since I wrote that code :smile:

zaggino:
sorry, can't review linux cpp code

appshell/appshell_extensions_gtk.cpp
```diff
@@ -758,6 +762,92 @@ static void Callback(GtkMenuItem* menuItem, gpointer userData) {
...
+            if (i != tokens.size() - 1) {
+                shortcut += "<";
+            }
+            shortcut += tokens[i];
+            if (i != tokens.size() - 1) {
+                shortcut += ">";
+            }
+        }
+        gchar* val = (gchar*)shortcut.c_str();
+        gtk_accelerator_parse (val, keyVal, modifier);
```
eyelash: nit: everywhere else there is no space between function name and parameters.

appshell/appshell_extensions_gtk.cpp
```diff
@@ -758,6 +762,92 @@ static void Callback(GtkMenuItem* menuItem, gpointer userData) {
...
+}
+
+
+int32 ParseShortcut(CefRefPtr<CefBrowser> browser, GtkWidget* entry, ExtensionString& key, GdkModifierType* modifier, guint* keyVal, ExtensionString& commandId) {
+    if (key.size()) {
+        GtkAccelGroup *accel_group = NULL;
+        accel_group = gtk_accel_group_new();
+
+        std::vector<std::string> tokens;
+        std::istringstream f(key.c_str());
```
eyelash: Is the `.c_str()` necessary here?

appshell/appshell_extensions_gtk.cpp
```diff
@@ -758,6 +762,92 @@ static void Callback(GtkMenuItem* menuItem, gpointer userData) {
...
+        std::string s;
+        while (getline(f, s, '-')) {
+            tokens.push_back(s);
```
eyelash: I'm wondering whether the tokens could be parsed here directly instead of copying them to a vector first and the parsing the vector.

appshell/appshell_extensions_gtk.cpp
```diff
@@ -758,6 +762,92 @@ static void Callback(GtkMenuItem* menuItem, gpointer userData) {
...
+    // we have to change it to "Ctrl+O". Careful not to change the final
+    // character, though, so "Ctrl--" ends up as "Ctrl+-".
+    if (result.size() > 0) {
+        replace(result.begin(), result.end() - 1, '-', '+');
+    }
+    return result;
+}
+
+
+int32 ParseShortcut(CefRefPtr<CefBrowser> browser, GtkWidget* entry, ExtensionString& key, GdkModifierType* modifier, guint* keyVal, ExtensionString& commandId) {
```
eyelash: This function could probably be static if it is not used outside this file.

appshell/appshell_extensions_gtk.cpp
saurabh95: Actually I have used tokens in creating shortcut of type <Ctrl>A and also in the fallback, so I won't be able to have fallback if there is no tokens vector.

saurabh95:
Thanks @eyelash for reviewing this PR :smile: 
I have updated the PR as per your comments.

eyelash:
@saurabh95 `GetDisplayKeyString` could probably be static as well. The rest looks fine to me. As I said, I only took a quick look at your code without compiling and I'm also no expert on GTK+ shortcuts.

appshell/appshell_extensions.cpp
```diff
@@ -606,6 +606,9 @@ class ProcessMessageDelegate : public ClientHandler::ProcessMessageDelegate {
                 bool enabled = argList->GetBool(2);
                 bool checked = argList->GetBool(3);
                 error = NativeMenuModel::getInstance(getMenuParent(browser)).setMenuItemState(command, enabled, checked);
+                if (error == NO_ERROR) {
+                    error = SetMenuItemState(browser, command, enabled, checked);
```
abhijitapte: this seems to be common code (i.e. windows, mac and win). 
Isn't the menu item state being set  in line 608?

saurabh95: Actually `NativeMenuModel::getInstance(getMenuParent(browser)).setMenuItemState(command, enabled, checked);` does not work for Linux as it only sets a flag, so I have added `SetMenuState` function which is meant only for Linux and for case of Windows and MAC it is No Op

ficristo:
I'm sorry, but I cannot review cpp code

appshell/appshell_extensions_gtk.cpp
```diff
@@ -652,69 +657,256 @@ int32 GetPendingFilesToOpen(ExtensionString& files)
...
+}
+
+static int GetPosition(const ExtensionString& positionString, const ExtensionString& relativeId,
+                       GtkWidget* container, NativeMenuModel& model)
+{
+    if (positionString == "" || positionString == "last")
+        return -1;
+    else if (positionString == "first")
+        return 0;
+    else if (relativeId.size() > 0) {
```
abhijitapte: use empty() instead

appshell/appshell_extensions_gtk.cpp
```diff
@@ -652,69 +657,256 @@ int32 GetPendingFilesToOpen(ExtensionString& files)
...
+        GList* iter;
+        for (iter = children; iter != NULL; iter = g_list_next(iter)) {
+            if (iter->data == relativeMenuItem)
+                break;
+            position++;
+        }
+        if (iter == NULL)
+            position = -1;
+        else if (positionString == "before")
+            ;
```
abhijitapte: can't you instead set position as 0 here.

saurabh95: We can't set position to 0 here, as in https://github.com/adobe/brackets-shell/pull/602/files#diff-10aa13d798996c2d964d38b1f328ae62R698 we are incrementing value of position, so setting it to 0 will always add the menu in the first index, which is not desired.

appshell/appshell_extensions_gtk.cpp
```diff
@@ -652,69 +657,256 @@ int32 GetPendingFilesToOpen(ExtensionString& files)
...
+        else if (positionString == "lastInSection") {
+            for (; iter != NULL; iter = g_list_next(iter)) {
+                if (GTK_IS_SEPARATOR_MENU_ITEM(iter->data))
+                    break;
+                position++;
+            }
+        }
+        else
+            position = -1;
+        g_list_free(children);
```
abhijitapte: can you point to the documentation in GTK where they specify that you need to explicitly free it?

appshell/appshell_extensions_gtk.cpp
```diff
@@ -652,69 +657,256 @@ int32 GetPendingFilesToOpen(ExtensionString& files)
...
+        return position;
+    }
+    else
+        return -1;
 }
 
 int32 AddMenu(CefRefPtr<CefBrowser> browser, ExtensionString title, ExtensionString command,
-              ExtensionString position, ExtensionString relativeId)
+              ExtensionString positionString, ExtensionString relativeId)
```
abhijitapte: why can't these parameters be passed by const ref?

appshell/appshell_extensions_gtk.cpp
```diff
@@ -652,69 +657,256 @@ int32 GetPendingFilesToOpen(ExtensionString& files)
...
+        std::vector<std::string> tokens;
+        std::istringstream f(key);
+        std::string s;
+        while (getline(f, s, '-')) {
+            tokens.push_back(s);
+        }
+        std::string shortcut = "";
+        // convert shortcut format
+        // e.g. Ctrl+A converts to <Ctrl>A whose entry is stored in accelerator table
+        for (int i=0;i<tokens.size();i++) {
```
abhijitapte: isn't it better to take the size() in a local and use that instead?

appshell/appshell_extensions_gtk.cpp
```diff
@@ -652,69 +657,256 @@ int32 GetPendingFilesToOpen(ExtensionString& files)
...
+            if (i != tokens.size() - 1) {
+                shortcut += "<";
+            }
+            shortcut += tokens[i];
+            if (i != tokens.size() - 1) {
```
abhijitapte: why is the last token excluded?

appshell/appshell_extensions_gtk.cpp
```diff
@@ -652,69 +657,256 @@ int32 GetPendingFilesToOpen(ExtensionString& files)
...
+            for (int i=0;i<tokens.size();i++) {
+                if (tokens[i] == "Ctrl") {
+                    modifier = GdkModifierType(modifier | GDK_CONTROL_MASK);
+                } else if (tokens[i] == "Alt") {
+                    modifier = GdkModifierType(modifier | GDK_MOD1_MASK);
+                } else if (tokens[i] == "Shift") {
+                    modifier = GdkModifierType(modifier | GDK_SHIFT_MASK);
+                } else if (tokens[i] == "Enter") {
+                    keyVal = GDK_KEY_KP_Enter;
+                } else if (tokens[i] == "Up" || tokens[i] == "\u2191") {
```
abhijitapte: It is better to use macros or constants instead of these string literals

appshell/appshell_extensions_gtk.cpp
saurabh95: These are actually implementation of interface which is used by all platforms, so changing this would mean changing in all the platforms implementation.
We can take this for later PR.

appshell/appshell_extensions_gtk.cpp
saurabh95: Actually in this we are changing the token string from Ctrl+A to <Ctrl>A whose entry is stored in accelerator table, so for the last token there is no need to enclose it in <>

appshell/appshell_extensions_gtk.cpp
saurabh95: https://developer.gnome.org/glib/stable/glib-Doubly-Linked-Lists.html#g-list-free

eyelash: also https://developer.gnome.org/gtk3/stable/GtkContainer.html#gtk-container-get-children says:

> **Returns** a newly-allocated list of the container’s non-internal children.

abhijitapte: Just checked the tooltip indeed conveys that we need to free.
I looked at documentation of https://developer.gnome.org/gtk3/stable/GtkContainer.html#gtk_container_get_focus_chain and noticed the explicit free mentioned there.

swmitra:
Great work @saurabh95 👍

eyelash:
I don't know if that was intentional but while merging this work to master all the commits were squashed (aa0336076aa485aed852af67f3194e970540a15c). Just to give you guys some feedback for the future: Seeing all your work merged without getting credit for it is not very motivating for a community contributor like me.

saurabh95:
@eyelash it was by mistake and we will be having Release Notes in which we highlight top features and their authors and we will have your name in that.
We will definitely take care of this from next time.

eyelash:
@saurabh95 Thank you, I'm glad to hear that.

PR #609, opened 2017-06-23
Linux UTF-8 BOM is preserved

saurabh95:


appshell/appshell_extensions.cpp
```diff
@@ -276,32 +276,36 @@ class ProcessMessageDelegate : public ClientHandler::ProcessMessageDelegate {
                 ExtensionString filename = argList->GetString(1);
                 ExtensionString encoding = argList->GetString(2);
                 std::string contents = "";
+                bool preserveBOM = false;
```
abhijitapte: why isn't a similar argument added in the shell api `appshell.fs.readFile`?

appshell/appshell_extensions_gtk.cpp
```diff
@@ -435,60 +435,53 @@ bool has_utf_32_BOM(gchar* data, gsize length)
...
+            if (detectedCharSet != "UTF-8") {
+                std::transform(detectedCharSet.begin(), detectedCharSet.end(), detectedCharSet.begin(), ::toupper);
+                DecodeContents(contents, detectedCharSet);
+                encoding = detectedCharSet;
+            }
+            else {
+                std::cout << encoding << "\n";
+                if (contents.length() >= 3 && contents[0] == (char)0xEF && contents[1] == (char)0xBB && contents[2] == (char)0xBF) {
+                    contents.erase(0,3);
+                    std::cout << "Preserve BOM" << "\n";
```
abhijitapte: need to remove this.

appshell/appshell_extensions_gtk.cpp
```diff
@@ -435,60 +435,53 @@ bool has_utf_32_BOM(gchar* data, gsize length)
...
+                detectedCharSet = encoding;
+            }
+            if (detectedCharSet != "UTF-8") {
+                std::transform(detectedCharSet.begin(), detectedCharSet.end(), detectedCharSet.begin(), ::toupper);
+                DecodeContents(contents, detectedCharSet);
+                encoding = detectedCharSet;
+            }
+            else {
+                std::cout << encoding << "\n";
+                if (contents.length() >= 3 && contents[0] == (char)0xEF && contents[1] == (char)0xBB && contents[2] == (char)0xBF) {
```
abhijitapte: could you use a macro instead and add proper comments as to why you have to do this check.

appshell/appshell_extensions_gtk.cpp
```diff
@@ -507,6 +500,8 @@ int32 WriteFile(ExtensionString filename, std::string contents, ExtensionString
         } catch (...) {
             error = ERR_CANT_READ;
         }
+    } else if (encoding == "UTF-8" && preserveBOM) {
+        contents = "\xEF\xBB\xBF" + contents;
```
abhijitapte: same here, need to add proper explanation of it as well as the string literal.

appshell/appshell_extensions.cpp
saurabh95: readfile is called from brackets and until then we don't know the encoding/bom of the file so default value is false.
so from brackets we don't need to send the argument

PR #613, opened 2017-06-29
Corrected UTF8 detection which was being detected as ISO-8859-1

saurabh95:


appshell/appshell_extensions_gtk.cpp
```diff
@@ -442,44 +442,54 @@ int32 ReadFile(ExtensionString filename, ExtensionString& encoding, std::string&
...
+        if (has_utf_32_BOM(file_get_contents, len)) {
+            error = ERR_UNSUPPORTED_ENCODING;
+        } else  if (has_utf8_BOM(file_get_contents, len)) {
+            std::ifstream file(filename.c_str());
+            std::stringstream ss;
+            ss << file.rdbuf();
+            contents = ss.str();
+            contents.erase(0, 3);
+            preserveBOM = true;
+        } else if (!g_locale_to_utf8(file_get_contents, -1, NULL, NULL, &gerror) || encoding != "UTF-8") {
```
saurabh95: Line 460 to 465 is the only change added new in this PR, rest is just revert of my last PR https://github.com/adobe/brackets-shell/pull/609/files#diff-10aa13d798996c2d964d38b1f328ae62

PR #620, opened 2017-08-09
Adding a check for node version during the build. 

harshit0405:
Adding a check for node version during the build. The setup task should fail if node version being bundled differs from the node version on the machine

### DIFF
--- a/tasks/setup.js
+++ b/tasks/setup.js
@@ -455,6 +455,33 @@ module.exports = function (grunt) {
         }
     });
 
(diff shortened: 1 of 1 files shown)


harshit0405:
@narayani28 Sorry, missed adding mac/linux support. Added that as well now. 

I've tested in on Windows and Mac. Couldn't test on a Linux machine as I don't have access to one currently.

saurabh95:
I have tested it on Linux. It is working fine.

PR #623, opened 2017-09-08
Null terminate the bracketsDirPath

boopeshmahendran:
This should fix issue https://github.com/adobe/brackets/issues/13661, https://github.com/adobe/brackets/issues/13681, https://github.com/adobe/brackets/issues/13682, https://github.com/adobe/brackets/issues/13680

@swmitra @nethip Please review.


nethip:
Thanks for fixing this @boopeshmahendran . It would be really helpful for documentation of your analysis (either here of in the code).

boopeshmahendran:
bracketsDirPath is used for constructing nodeExecutablePath and nodecorePath. But bracketsDirPath is not null terminated. So nodeExecutablePath and nodeCorePath are wrong and node is not started.

nethip:
@boopeshmahendran Thanks for fixing this.

PR #626, opened 2017-12-12
[Linux only] CEF Upgrade to 2785

nethip:
CEF needs to be upgraded to a later version, as the current version of CEF has some issues with menus text getting corrupt. Also CEF needs to be patched wit IME as IME handling is missing in CEF. 

The following are the note worthy things.

1) Upgraded to CEF 2785
2) Moved the initialization of brackets object to `Global.js`
3) ~~Forced the DPI scale factor to 1, as rendering is breaking upon launching. Surprisingly, upon resizing,  the view starts to render properly. We  need to revisit this in the next sprint.~~(This is not required anymore, as the rendering issue is now fixed with a different technique with a fake resize)

### DIFF
--- a/appshell/cefclient_gtk.cc
+++ b/appshell/cefclient_gtk.cc
@@ -31,7 +31,6 @@
 static std::string APPICONS[] = {"appshell32.png","appshell48.png","appshell128.png","appshell256.png"};
 int add_handler_id;
 bool isReallyClosing = false;
-char *DEVICE_SCALE_COMMAND_LINE_OPTION = "--force-device-scale-factor=1";
 
 // The global ClientHandler reference.
 extern CefRefPtr<ClientHandler> g_handler;
--- a/appshell/cefclient_gtk.cc
+++ b/appshell/cefclient_gtk.cc
@@ -112,10 +112,8 @@ void TerminationSignalHandler(int signatl) {
 }
 
 int RunMain(int argc, char* argv[]) {
-
   // Create a copy of |argv| on Linux because Chromium mangles the value
   // internally (see issue #620).
-
   CefScopedArgArray scoped_arg_array(argc, argv);
   char** argv_copy = scoped_arg_array.array();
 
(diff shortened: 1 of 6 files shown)


nethip:
Brackets relevant PR https://github.com/adobe/brackets/pull/13975. Both need to be merged.

nethip:
Relvant https://github.com/adobe/brackets/pull/13975

nethip:
@abhijitapte Could you please review this?

appshell/cefclient_gtk.cc
```diff
@@ -111,9 +111,11 @@ void TerminationSignalHandler(int signatl) {
   MainContext::Get()->GetRootWindowManager()->CloseAllWindows(true);
 }
 
-int RunMain(int argc, char* argv[]) {
+int RunMain(int argc, char** argv) {
```
abhijitapte: might not be really needed in this PR's context

navch:
Please keep track of 3rd point to pick that up in next sprint. Apart from that it looks good to me 👍

appshell/client_handler.cpp
```diff
@@ -202,7 +202,11 @@ bool ClientHandler::OnDragEnter(CefRefPtr<CefBrowser> browser,
 }
 
 void ClientHandler::OnLoadStart(CefRefPtr<CefBrowser> browser,
-                                CefRefPtr<CefFrame> frame) {
+                                CefRefPtr<CefFrame> frame
+                                #ifdef OS_LINUX
```
abhijitapte: is the function's signature like this for linux plaform alone?

appshell/cefclient_gtk.cc
nethip: @abhijitapte I will revert this change.

appshell/client_handler.cpp
nethip: This is required as we have migrated to a newer version of CEF and the signature changed in the newer version.

nethip:
@navch The third option is no more required as the issue was fixed in a different way.

appshell/client_handler.cpp
abhijitapte: got you. this will have to be removed after we are on 2785 on all three platforms.

nethip:
@abhijitapte Addressed the review comments. Please have a look at the PR.

PR #629, opened 2018-01-30
Merge release back to master

ficristo:
Do not squash the commits.


ficristo:
Given the conflicts this has to be done from the command line.
@nethip could you do it?

nit: following semantic versioning maybe instead of `product.release.number.maint` should be `product.release.number.patch` ?

PR #630, opened 2018-02-01
Restoring the brackets alias intitialization in appshell_extensions.js

nethip:
This is still going to be required as brackets.app.language was getting accessed even before it was getting initialized in global.js. 

I have unit tested on Win, Linux and MAC with this change and the change seems to be working fine. Also I tested by hard routing the language to "fr", and Brackets is now correctly able to take the language, defined by `brackets.app.language`.

Thanks @swmitra for debugging this. Could you please review this?
@abhijitapte It would be great if you can review this as well.

### DIFF
--- a/appshell/appshell_extensions.js
+++ b/appshell/appshell_extensions.js
@@ -27,7 +27,7 @@
 // Note: All file native file i/o functions are synchronous, but are exposed
 // here as asynchronous calls. 
 
-var appshell;
+var appshell, brackets;
 if (!appshell) {
     appshell = {};
 }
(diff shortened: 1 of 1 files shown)


swmitra:
@nethip A very minor change just for better readability purpose. 
If we can define the alias "brackets" upfront while defining "appshell" and then assign the initialised "appshell" placeholder to the alias before going through the self executing member initialisation block.

nethip:
@swmitra Thanks for reviewing this! I will update the PR with your review comments.

PR #636, opened 2018-04-07
Linux window size and position 

pelatx:
In the different Linux versions (Debian 8 and 9, Ubuntu 16) in which I have tested Brackets compiled from the Master branch with the latest commits (same result in the lastest release), this still does not remember the position and size of the window from one execution to another (issues #13182 and #7174).

I have fixed this using a "key file" that is created automatically by Brackets, updated every time the application is closed and read every time it starts.

Maybe it's not the best solution, but I've been using it for a few days and it works well. So I leave this PR for your consideration.

Thank you.

Edited: tested in the three distributions mentioned above.

### DIFF
--- a/appshell/browser/root_window_gtk.cc
+++ b/appshell/browser/root_window_gtk.cc
@@ -78,6 +78,13 @@ void SaveWindowState(GtkWindow* window) {
   gint width  = DEFAULT_WINDOW_WIDTH;
   gint height = DEFAULT_WINDOW_HEIGHT;
 
+  // Try to center the window.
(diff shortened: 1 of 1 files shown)


nethip:
@pelatx Thanks for this PR. I have raised a PR on your fork with some code cleanup. Can you have a look at this PR https://github.com/pelatx/brackets-shell/pull/1 and merge it into your branch, if changes are looking fine.

pelatx:
Thanks @nethip.

nethip:
@pelatx Thanks for this PR!

PR #637, opened 2018-04-11
Linux native menus lost shortcuts

pelatx:
This PR makes the shortcuts to be shown again in the Linux native menus. Which were lost in my previous PR relative to the switchable menu entries.

A few shortcuts are still not shown. And honestly, at the moment I could not find the reason. Maybe someone can point out what I've missed.

The latter is more the reason for opening this PR than a merge in this state.

Thank you.

### DIFF
--- a/appshell/appshell_extensions_gtk.cpp
+++ b/appshell/appshell_extensions_gtk.cpp
@@ -1064,6 +1064,7 @@ int32 AddMenuItem(CefRefPtr<CefBrowser> browser, ExtensionString parentCommand, 
 
     ExtensionString commandId = model.getCommandId(tag);
     model.setOsItem(tag, entry);
+	model.setKey(tag, key);
     ParseShortcut(browser, entry, key, commandId);
     GtkWidget* menuHeader = (GtkWidget*) model.getOsItem(parentTag);
     GtkWidget* menuWidget = gtk_menu_item_get_submenu(GTK_MENU_ITEM(menuHeader));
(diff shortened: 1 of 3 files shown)


saurabh95:
Hey, @pelatx Could you please point to menu items for which shortcuts are not visible. It would be helpful for us in process of finding the reason for it.
Also, I would recommend that you install emmet extension and check for menu items under emmet menu. I used that extension only for my testing since it has a lot of menu items (checked + non-checked) when I worked adding checked menu items.

pelatx:
Sorry for the delay.

I've been doing a little more research and I've seen where the problem is coming from.

The menu entries that configure the shortcut when added to the menu ( in `AddMenuItem()`) are those that work as expected.
The problem occurs when this is not done in this way and the shortcut is then configured by calling `SetMenuItemShortcut()`.

I have tried to solve it by setting the key in the model (`model.setKey(tag)`) within `SetMenuItemShortcut()`. And then recover it in `SetMenuItemState()` to pass it to `ParseShortcut()` within.

It does not work for now. I will keep trying.

PS: I do not know if I explained it well. :sweat_smile:

saurabh95:
Thanks @pelatx, this will surely help us in finding the probable solution.

Your solution seems correct to me, not sure why it is not working. Did you check if all the keys are stored when using `SetMenuItemShortcut `, I mean could you please check by logging all the keys present in the model when it reaches `SetMenuItemShortcut ` and verify if all the desired keys are present or not? This will give us an idea if we are updating the correct map. 

PS: This is just a hunch, I have not tried this myself.

pelatx:
Right hunch, @saurabh95.

I do not know how I could look so many times at the `SetMenuItemShortcut` function without seeing the '&' missing in `NativeMenuModel & model = ...`

saurabh95:
@nethip Could you please review the changes? 
It is LGTM from my side.

nethip:
Thanks for the fix @pelatx! LGTM. Is it OK If I squash the commits and merge?

pelatx:
It is OK @nethip.

nethip:
Thanks @pelatx and thanks @saurabh95 for reviewing this as well.

PR #638, opened 2018-04-13
Fix full screen window issue in multiple montiors

sobisht:
Issue: https://github.com/adobe/brackets/issues/13349
- Currently we are not handling the WM_NCPAINT for dark aero glass. This causes the issue when we delegate this handling to the OS.
To fix this issue, I have added a handle for dark aero similar to the one present for dark window.

- This will also fix the following issue 
brackets

The boundaries of the non client area is not correctly painted when we launch brackets in maximized window and go from maximize to restored window.
After the fix 
brackets_fix
 
Unit tested the fix 
- On Win10 builds before and after creators update.
- On Dual monitors with different resolution
- On Dual monitors with one monitor with high DPI.

Please review @nethip @swmitra @vickramdhawal @JeffryBooher

### DIFF
--- a/appshell/cef_dark_aero_window.cpp
+++ b/appshell/cef_dark_aero_window.cpp
@@ -843,6 +843,10 @@ LRESULT cef_dark_aero_window::WindowProc(UINT message, WPARAM wParam, LPARAM lPa
                 return 0L;
         }
         break;
+	case WM_NCPAINT:
+		if (HandleNcPaint((HRGN)wParam))
+			return 0L;
+		break;
     }
 
     // call DefWindowProc?


nethip:
Thanks for the PR @sobisht . Great work!

I see that these are also not handled inside `cef_aero_dark_window`. Does it make sense to handle the following as well?

```
WM_NCCREATE
WM_NCDESTROY
WM_NCHITTEST
```

nethip:
Thanks for confirming @sobisht!

nethip:
Thanks for fixing this @sobisht

PR #639, opened 2018-04-20
AutoUpdate Feature Health Data Logging Changes

niteskum:
This PR includes changes for Auto Update Feature Health Data Logging Changes

### DIFF
--- a/installer/mac/update.sh
+++ b/installer/mac/update.sh
@@ -38,37 +38,37 @@ function printError(){
 function verifySetup(){
     if [ "$APP_DIR" == "$EMPTY_STRING" ]
     then
-        printError "Brackets app dir is missing"
+        printError "Brackets app dir is missing: INS_O1"
         return $OPTIONS_MISSING
     fi
 
     if [ "$APP_NAME" == "$EMPTY_STRING" ]
     then
-        printError "Brackets app is missing"
+        printError "Brackets app is missing: INS_02"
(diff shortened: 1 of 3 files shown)


navch:
Can you please add some description, why we doing this?

installer/mac/update.sh
```diff
@@ -176,13 +176,13 @@ function updateBrackets(){
         exitStatus=$?
         if [ $exitStatus -ne 0 ]
         then
-            printError "Brackets could not be opened. open returnd $exitStatus."
+            printError "Brackets could not be opened. open returnd $exitStatus : INS_12"
         else
             printInfo "Opened Brackets."
-        fi
+        fiS
```
navch: Is this intentional?

installer/mac/update.sh
```diff
@@ -38,37 +38,37 @@ function printError(){
 function verifySetup(){
     if [ "$APP_DIR" == "$EMPTY_STRING" ]
     then
-        printError "Brackets app dir is missing"
+        printError "Brackets app dir is missing: INS_O1"
```
navch: Don't we need "Error" word in these strings?

niteskum: "ERROR:" String will be added by printError Function so we don't need to add explicitly.

nethip:
Thanks @niteskum for the changes.

PR #640, opened 2018-04-24
Adding maint version to brackets win installer

gavankar:
This will allow an prerelease installer to upgrade a release build
easily.


charlieantao:
LGTM

swmitra:
@gavankar The change set looks good to me, but I do have one question - 

> Can't we bring in `build number` to identify a release internally? Let the version string act like a `label` for the version, but all comparisons should be done based on `build number`. Otherwise we are left with same problem if we refresh a build...

gavankar:
@swmitra you are right that build number would be better. But i could not find any variable/property from which to get the build number while grunt is creating the installer.

PR #641, opened 2018-04-25
AutoUpdate Windows Installer Command changed from Verbose to Info Logging

niteskum:


niteskum:
in This PR Auto Update Windows Installer Command  has been changed to Info Logging.
Which reduces Installer Log Size from 50 MBs to few KBs

niteskum:
ping @mbhavi @navch @swmitra for review

appshell/update_win.cpp
```diff
@@ -54,7 +54,7 @@ void UpdateHelper::RunAppUpdate() {
 		std::wstring commandInput = L"msiexec /i " + m_InstallerPath + L" /qr";
 		//AUTOUPDATE_PRERELEASE
 		if (!m_logFilePath.empty()) {
-			commandInput += L" /l*V " + m_logFilePath;
+			commandInput += L" /li " + m_logFilePath;
```
mbhavi: @niteskum , please verify the success or failure log message in the logs, if it changes because of this change.
Also, roughly, what is the impact of this change on the performance of updation?

niteskum: There in no change in Log Error or Success Message.
Performance improvement in installation is of around 10 secs.
But Log Size has reduced to 15 KB from 50 MB.

niteskum:
Charles Antao recommended to use below Installer command 
msiexec /i  <installerFilePath> /qr /li <logFilePath> LAUNCH_APPLICATION_SILENT=1
 
Earlier it was using The /*v option  which results in extremely verbose logging and that might  cause the slowdown.
in this PR I have changed the Installer command to use option "/li".
ping @swmitra for review

navch:
Bhavika has reviewed this PR. This is just limited to installer logs. So merging this PR.

PR #642, opened 2018-04-27
Disable custom drag handlers if OS is linux

boopeshmahendran:
This PR disables custom drag handlers to make CEF fallback to default behavior

### DIFF
--- a/appshell/browser/client_handler.cc
+++ b/appshell/browser/client_handler.cc
@@ -323,6 +323,7 @@ void ClientHandler::OnDownloadUpdated(
   }
 }
 
+#ifndef OS_LINUX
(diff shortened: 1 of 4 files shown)


vickramdhawal:
@boopeshmahendran Can you put in more details in the PR to understand how this change helps.

boopeshmahendran:
Before this change, on Linux, the events related to drag and drop such as dragover and drop were not triggered on the brackets side and as a result, the drag and drop feature in file tree was not working.
This changes fixes that issue by disabling the custom drag and drop handling and making CEF fallback to the default behavior.

vickramdhawal:
LGTM 👍

PR #647, opened 2018-05-11
Disable GPU for Ubuntu18.04

gautam0217:


appshell/cefclient_gtk.cc
```diff
@@ -233,5 +233,8 @@ int RunMain(int argc, char* argv[]) {
 
 // Program entry point function.
 int main(int argc, char* argv[]) {
+	char extra_option[14];
```
shubhsnov: The indendation is off.

appshell/cefclient_gtk.cc
```diff
@@ -233,5 +233,8 @@ int RunMain(int argc, char* argv[]) {
...
+	snprintf(extra_option,sizeof(extra_option),"--disable-gpu");
```
shubhsnov: Instead of doing all this, why don't you use std:string, it's cleaner, and bounds safe.

Something like: 
```
std::string gpuArg = "--disable-gpu";
argv[argc] = gpuArgs.c_str();
argc++;
```

gautam0217: I don't want to go to heap.

shubhsnov: I don't get your point. The optimization(vendor dependent) here won't matter since this is a very small operation at start up, and if it does happen, it will only be for good. Moreover what I wrote came from the general conventions followed in brackets-shell repo. See usage of CefString and std::string. 

Also the indentation and spacing in the code is still off.

PR #649, opened 2018-05-13
Build fixes for Ubuntu18.04

gautam0217:


gautam0217:
Link to old review https://github.com/adobe/brackets-shell/pull/645

appshell/appshell_extensions_platform.cpp
```diff
@@ -7,6 +7,7 @@
 #include "appshell/browser/main_context.h"
 #include "appshell/browser/root_window_manager.h"
 #include "appshell/browser/root_window_gtk.h"
+#include <unicode/unistr.h>
```
shubhsnov: Do these headers work for all linux versions?

gautam0217: Yes, I have just resolved indirect header dependency in this case.
These are standard unicode library headers.

PR #651, opened 2018-05-23
Fix Issue 14149 - Fixes Dns rebinding attack in CEF

sabanaya:
Fixes [#14149](https://github.com/adobe/brackets/issues/14149)
Applied the CEF patch on Linux and Windows and created the distribution.
Upgraded CEF from 2623 to 2704 on Mac and applied patch.

@abhijitapte @vickramdhawal  @nethip @swmitra Please review


vickramdhawal:
@swmitra I think this is good to merge as well. what do you think ?

PR #652, opened 2018-05-25
Close file handle on unsupported encoding

boopeshmahendran:
Fixes [#14244](https://github.com/adobe/brackets/issues/14244)

For some [specific](https://github.com/adobe/brackets-shell/blob/master/appshell/appshell_extensions_win.cpp#L971) encodings, the **file handle is not closed** which causes error code 32 (The process cannot access the file as it is used by some other process) while calling the MoveFile system function.

@vickramdhawal @swmitra Please review

### DIFF
--- a/appshell/appshell_extensions_win.cpp
+++ b/appshell/appshell_extensions_win.cpp
@@ -969,7 +969,7 @@ int32 ReadFile(ExtensionString filename, ExtensionString& encoding, std::string&
                                 detectedCharSet = conv.to_bytes(encoding);
                             }
                             if (detectedCharSet == "UTF-16LE" || detectedCharSet == "UTF-16BE") {
-								CloseHandle(hFile);
+				 CloseHandle(hFile);
                                 return ERR_UNSUPPORTED_UTF16_ENCODING;
                             }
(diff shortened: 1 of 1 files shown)


appshell/appshell_extensions_win.cpp
```diff
@@ -969,6 +969,7 @@ int32 ReadFile(ExtensionString filename, ExtensionString& encoding, std::string&
                                 detectedCharSet = conv.to_bytes(encoding);
                             }
                             if (detectedCharSet == "UTF-16LE" || detectedCharSet == "UTF-16BE") {
+				 CloseHandle(hFile);
```
vickramdhawal: this looks good. Are there other places where File Handles need to be closed ? Like in catch blocks ?

vickramdhawal:
good job on resolving this @boopeshmahendran 👍

appshell/appshell_extensions_win.cpp
boopeshmahendran: I think this is the only place. FYI, this implementation was changed recently in this [PR](https://github.com/adobe/brackets-shell/pull/635/files).

boopeshmahendran:
Thank you very much for reviewing @saurabh95.

PR #653, opened 2018-06-01
Upgrading node to v6.14.0

subhashjha333:


vickramdhawal:
Hi @subhashjha333 . Thanks for contributing. Can you describe why this changes is necessary and how the minor version bump is benefitting ?

subhashjha333:
@vickramdhawal  The customer have reported few security issues in Dreamweaver in current version of node which has been fixed in v6.12.2. On the top of that node has some additional fixes in v6.14.0. Thus I picked up this for integration.

v6.11.0  contains the vulnerability CVE-2017-15896 as described in https://nodejs.org/en/blog/vulnerability/december-2017-security-releases/.

nethip:
@subhashjha333 Thanks for this PR!

PR #660, opened 2018-10-15
Forcefully disabling renderer accesibility

nethip:
Windows update KB4343909 is leading to performance issues on some systems. This is because of accessiblity renderer tree getting created wrongly. So disabling this for now.

Thanks to @katoken-0215 for pointing this out.

@abhijitapte @swmitra Could you please review this?

### DIFF
--- a/appshell/cefclient.cpp
+++ b/appshell/cefclient.cpp
@@ -19,7 +19,10 @@
 #include "config.h"
 
 CefRefPtr<ClientHandler> g_handler;
-bool					 g_force_enable_acc = false;
+
+#ifdef OS_WIN
+bool g_force_enable_acc = false;
+#endif
 
 CefRefPtr<CefBrowser> AppGetBrowser() {
   if (!g_handler.get())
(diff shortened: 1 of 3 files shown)


nethip:
Fixes https://github.com/adobe/brackets/issues/14515

appshell/cefclient.cpp
```diff
@@ -32,6 +33,20 @@ CefWindowHandle AppGetMainHwnd() {
   return g_handler->GetMainHwnd();
 }
 
+// CefCommandLine::HasSwitch is unable to report the presense of switches,
```
vickramdhawal: i'm not too sure about this. can you verify with "force-renderer-accessibility" without the leading "--" and use the existing method.

nethip: @vickramdhawal I tried doing searching for the switch without the — . It is not working. However when I search for the same switch in ‘ClientApp::OnBeforeCommandLineProcessing’ it is being found. Again the same is not found in ‘ClientApp::OnBeforeChildProcessLaunch’. Something weird going on there.

vickramdhawal: If this is confirmed, then the changes LGTM :+1

PR #661, opened 2018-10-16
Cef 3538 integration

gautam0217:


gautam0217:
I am still not able to enable paste contents from clipboard in `debug > New Brackets Window`.
First Brackets window is working correctly but `new window` opened from `debug > New Brackets Window` does not have paste functionality.

shubhsnov:
Closing this as we plan to move to 3638 on all platforms.

PR #662, opened 2018-12-05
FIx for #14171. libcurl dependency issue.

PlamenSh:
Details here: https://github.com/adobe/brackets/issues/14171
In short libcurl3 dependency causing problems on linux with libcurl4.


xekon:
This seems to still be an issue, today Atom code editor started randomly crashing.

I decided to give Brackets a go, only to find out that it depends on libcurl3 but i have libcurl4 installed.

Apparently this could have been fixed years ago...

Is the Brackets project just dead in the water or what?

PR #664, opened 2019-04-02
Win64 build fix

gautam0217:
Windows 64 bit build can be done using following commands. I am using this build as my default brackets and have not seen any issue yet.
TODO: installer is not working.
```
grunt setup --force-target-word-size=64 --verbose
grunt build  --force-target-word-size=64 --verbose
```

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -91,23 +91,43 @@ module.exports = function (grunt) {
                 "src"       :  "<%= icu.url %>/icu_<%= icu.version %>_macosx64.zip"
             },
             /* win */
-            "cef-win": {
+            "cef-win32": {
                 "dest"      : "<%= downloads %>",
                 "src"       : "<%= cef.url %>/cef_binary_<%= cef.version %>_windows32.zip"
             },
-            "cef-win-symbols": {
+            "cef-win32-symbols": {
                 "src"  : ["<%= cef.url %>/cef_binary_<%= cef.version %>_windows32_debug_symbols.zip", "<%= cef.url %>/cef_binary_<%= cef.version %>_windows32_release_symbols.zip"],
                 "dest" : "<%= downloads %>/cefsymbols"
             },
-            "node-win": {
+            "node-win32": {
                 "dest"      : "<%= downloads %>",
                 "src"       : "http://nodejs.org/dist/v<%= node.version %>/win-x86/node.exe"
             },
-            "icu-win": {
+            "icu-win32": {
                 "dest"      :  "<%= downloads %>",
                 "src"       :  "<%= icu.url %>/icu_<%= icu.version %>_windows32.zip"
             },
-            "vs-crt-win": {
+            "vs-crt-win32": {
+                "dest"      :  "<%= downloads %>",
+                "src"       :  "<%= vsCrt.url %>/vs<%= vsCrt.version %>-crt-ia32.zip"
+            },
+            "cef-win64": {
+                "dest"      : "<%= downloads %>",
+                "src"       : "<%= cef.url %>/cef_binary_<%= cef.version %>_windows64.zip"
+            },
+            "cef-win64-symbols": {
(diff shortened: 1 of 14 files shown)


shubhsnov:
Nit: Change the command line flag:  to `--arch=x64` instead of `--force-target-word-size=64`

gautam0217:
On windows it common.arch()  returns empty on windows, so added option to enable 64 bit build while 32 bit is default.
so, `--archbit=64` looks more suitable to me.

gautam0217:
Addressed review comments.

Gruntfile.js
```diff
@@ -91,23 +91,43 @@ module.exports = function (grunt) {
...
+                "dest"      : "<%= downloads %>",
+                "src"       : "http://nodejs.org/dist/v<%= node.version %>/win-x64/node.exe"
+            },
+            "icu-win64": {
+                "dest"      :  "<%= downloads %>",
+                "src"       :  "<%= icu.url %>/icu_<%= icu.version %>_windows64.zip"
+            },
+            "vs-crt-win64": {
                 "dest"      :  "<%= downloads %>",
                 "src"       :  "<%= vsCrt.url %>/vs<%= vsCrt.version %>-crt-ia32.zip"
```
vickramdhawal: Is there a 64bit equivalent available to be used here ?

appshell/cef_window.h
```diff
@@ -249,3 +272,17 @@ class cef_window
 
     BOOL TrackNonClientMouseEvents(bool track = true);
 };
+
+
+#include <type_traits>
+#include <cstring>
+namespace from_cpp20 {
+	template<typename Dest, typename Source>
+		inline auto bit_cast(Source const & src) -> typename std::enable_if<sizeof(Source) == sizeof(Dest), Dest>::type
```
vickramdhawal: there are a couple of more conditions to the enable_if in the std code https://en.cppreference.com/w/cpp/numeric/bit_cast see if it makes sense to incorporate that.

tasks/common.js
```diff
@@ -233,7 +233,12 @@ module.exports = function (grunt) {
         return _platform;
     }
     
-    function arch() {
+    function arch(hint) {
+		hint = !hint && platform() === 'win' ? 'ia32' : hint;
```
vickramdhawal: we can add platform() === 'win' || platform() === "linux " and then it will be possible to remove the code below added for linux.

appshell/cef_window.h
gautam0217: I am not looking to implement a full proof back porting of C++20 `bit_cast`.
I can restrict it's uses further by asserting for pointer using `std::pointer_traits` or `std::is_pointer`.

tasks/common.js
gautam0217: Done.

tasks/common.js
```diff
@@ -233,15 +233,12 @@ module.exports = function (grunt) {
...
-            if (process.arch === "x64") {
-                return 64;
-            } else {
-                return 32;
-            }
-        }
-        
+    function arch(hint) {
+		hint = hint ? hint : process.arch;
+		hint = platform() === 'mac' ? undefined : hint;
```
vickramdhawal: In case of Mac we are returning "" - I'm guessing that is expected.
hint = (platform() !== 'mac' && hint) ? hint : ""; is more concise and precise.

tasks/build.js
```diff
@@ -59,9 +59,12 @@ module.exports = function (grunt) {
     grunt.registerTask("full-build", ["git", "create-project", "build-www", "build", "stage", "package"]);
     grunt.registerTask("installer", ["set-release-optional", "full-build", "build-installer"]);
 
+    //Not to be used for MacOS
+    var target_arch_hint = grunt.option('arch') || undefined;
```
shubhsnov: grunt.option should return undefined in case no args are passed, so this is superfluous.

tasks/common.js
shubhsnov: Shouldn't we just return with "" in case of mac. Linux and Win will have similar behaviour after that.

Gruntfile.js
gautam0217: Yes, started using 64 bit vs-crt.

gautam0217:
@vickramdhawal addressed all the review comments.

PR #666, opened 2019-05-01
CEF 3683 on Ubuntu 18.04 Mate desktop

gautam0217:
Basic brackets functionalities are working my local testing.
Only missing thing is copy-paste is disable on Debug > New Brackets Window.
Copy paste is working on all Debug windows opened.

CEF downloaded from [soptify](http://opensource.spotify.com/cefbuilds/index.html.)
ICU58 build from source using following config:
```
export CFLAGS="-fPIC" && export CXXFLAGS="-fPIC" && ./configure --prefix=/home/gautam/icu_58_linux64_release --enable-static
```


gautam0217:
binaries used to build are: [ICU](https://drive.google.com/open?id=1VvJaa3Zsfhx6q1sznn8259CTftjleEY1) and [CEF](https://drive.google.com/open?id=1R2mtleRYc_w8jmb0k1Kf77sQ7G8lYynr)

appshell/cefclient.cpp
```diff
@@ -28,7 +28,7 @@ CefRefPtr<CefBrowser> AppGetBrowser() {
 
 CefWindowHandle AppGetMainHwnd() {
   if (!g_handler.get())
-    return NULL;
+    return 0;
```
shubhsnov: Why are we using an explicit integer for pointer here? CefWindowHandle is defined as HWND on Windows, NSView* on Mac OS X and GtkWidget* on Linux.

`NULL` seemed like a better choice. `nullptr` even better to handle type ambiguity with int.

appshell/client_handler.cpp
```diff
@@ -280,65 +278,6 @@ void ClientHandler::OnLoadingStateChange(CefRefPtr<CefBrowser> browser,
   SetNavState(canGoBack, canGoForward);
 }
 
-bool ClientHandler::OnConsoleMessage(CefRefPtr<CefBrowser> browser,
```
shubhsnov: Some changes here are not limited to Linux. I would recommend Linux only changes for this PR.

shubhsnov:
@gautam0217 you also need to change: https://github.com/adobe/brackets-shell/blob/master/installer/linux/debian/control 

The `Depends` key should reflect the Depends of the Chrome equivalent of CEF 3683.

`dpkg -p {package}` should give info of the chrome package as well.

PR #668, opened 2019-11-06
Disabled remote-debugging by default and appshell.app.getRemoteDebuggingPort() needs a callback argument to work

jha-g:
* Disabling remote debugging by default, and it can be enabled by passing a command line arg as --remote-debugging-port=xxxx
* Changed appshell.app.getRemoteDebuggingPort() implementation to require a callback
* Added validation for port range, and also without remote-debugging-port we will not be able to debug brackets using a browser, and if browser debugging is disabled, appshell.app.getRemoteDebuggingPort(callback) will print 0

### DIFF
--- a/appshell/appshell_extensions.cpp
+++ b/appshell/appshell_extensions.cpp
@@ -842,6 +842,8 @@ public:
             uberDict->SetList(0, dirContents);
             uberDict->SetList(1, allStats);
             responseArgs->SetList(2, uberDict);
+        } else if (message_name == "GetRemoteDebuggingPort") {
+            responseArgs->SetInt(2, g_remote_debugging_port);
         }
 
         else {
--- a/appshell/appshell_extensions.cpp
+++ b/appshell/appshell_extensions.cpp
@@ -37,6 +37,7 @@
 #include "update.h"
 
 extern std::vector<CefString> gDroppedFiles;
+extern int g_remote_debugging_port;
 
 namespace appshell_extensions {
 
--- a/appshell/cefclient.cpp
+++ b/appshell/cefclient.cpp
@@ -7,7 +7,6 @@
 #include <cstdlib>
 #include <sstream>
 #include <string>
-#include <limits>
 #include "include/cef_app.h"
 #include "include/cef_browser.h"
 #include "include/cef_command_line.h"
@@ -24,7 +23,6 @@ int g_remote_debugging_port = 0;
 
 #ifdef OS_WIN
 bool g_force_enable_acc = false;
-#undef max
 #endif
 
 CefRefPtr<CefBrowser> AppGetBrowser() {
@@ -101,14 +99,16 @@ void AppGetSettings(CefSettings& settings, CefRefPtr<CefCommandLine> command_lin
   CefString debugger_port = command_line->GetSwitchValue("remote-debugging-port");
   if (!debugger_port.empty()) {
     int port = atoi(debugger_port.ToString().c_str());
-    static const int max_port_num = static_cast<int>(std::numeric_limits<uint16_t>::max());
-    if (port > 1024 && port < max_port_num) {
+    static const int max_port_num = 65535;
+    static const int max_reserved_port_num = 1024;
+    if (port > max_reserved_port_num && port < max_port_num) {
       g_remote_debugging_port = port;
       settings.remote_debugging_port = port;
     }
     else {
       LOG(ERROR) << "Could not enable remote debugging on port: "<< port
-                 << "; port number must be greater than 1024 and less than " << max_port_num;
+                 << "; port number must be greater than "<< max_reserved_port_num
+                 << "and less than " << max_port_num;
     }
   }
   
--- a/appshell/cefclient.cpp
+++ b/appshell/cefclient.cpp
@@ -7,6 +7,7 @@
 #include <cstdlib>
 #include <sstream>
 #include <string>
+#include <limits>
 #include "include/cef_app.h"
 #include "include/cef_browser.h"
 #include "include/cef_command_line.h"
@@ -19,9 +20,11 @@
 #include "config.h"
 
 CefRefPtr<ClientHandler> g_handler;
+int g_remote_debugging_port = 0;
 
 #ifdef OS_WIN
 bool g_force_enable_acc = false;
+#undef max
 #endif
 
 CefRefPtr<CefBrowser> AppGetBrowser() {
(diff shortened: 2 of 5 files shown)


appshell/cefclient.cpp
```diff
@@ -95,7 +98,19 @@ void AppGetSettings(CefSettings& settings, CefRefPtr<CefCommandLine> command_lin
       command_line->GetSwitchValue(client::switches::kJavascriptFlags);
     
   // Enable dev tools
-  settings.remote_debugging_port = REMOTE_DEBUGGING_PORT;
+  CefString debugger_port = command_line->GetSwitchValue("remote-debugging-port");
+  if (!debugger_port.empty()) {
+    int port = atoi(debugger_port.ToString().c_str());
+    static const int max_port_num = static_cast<int>(std::numeric_limits<uint16_t>::max());
```
shubhsnov: What happens if the port is already in use?

appshell/appshell_extensions.cpp
```diff
@@ -842,6 +842,8 @@ class ProcessMessageDelegate : public ClientHandler::ProcessMessageDelegate {
             uberDict->SetList(0, dirContents);
             uberDict->SetList(1, allStats);
             responseArgs->SetList(2, uberDict);
+        } else if (message_name == "GetRemoteDebuggingPort") {
```
shubhsnov: We're changing the normal function to one that takes callback? Does this have any impact on existing usages? This is a breaking change, though I guess no extensions will ideally use this.

appshell/cefclient.cpp
jha-g: I checked with `python -m SimpleHTTPServer 9999` and Brackets snatches the port from Python HTTP Server.

shubhsnov: Can we get away without needing limits header? 
I am not comfortable with the need for unsetting standard functions.
You can compute uint16 max as such, will work for all platforms
uint16 max_limit = 0;
max_limit = ~max_limit;
max_limit = max_limit >> 1

jha-g: std::numeric_limits<> is C++ equivalent of C macros denoting MAX and MIN of a default numeric data types. So, instead of `std::numeric_limits<uint16_t>::max()`, I could use `UINT16_MAX` defined in [stdint.h](https://en.cppreference.com/w/cpp/header/cstdint)

appshell/appshell_extensions.cpp
jha-g: Yes, it does impact one use case, and fixed that with commit: https://github.com/adobe/brackets/pull/14956/commits/8677756aa443b98523852f86bfcbfc1a77d93478

appshell/cefclient.cpp
jha-g: Don't bother about header, it's C++, not JS.

jha-g:
Addressed all the review comments.

swmitra:
@vickramdhawal Can you please take a quick look?

appshell/cefclient.cpp
```diff
@@ -95,7 +96,21 @@ void AppGetSettings(CefSettings& settings, CefRefPtr<CefCommandLine> command_lin
       command_line->GetSwitchValue(client::switches::kJavascriptFlags);
     
   // Enable dev tools
-  settings.remote_debugging_port = REMOTE_DEBUGGING_PORT;
+  CefString debugger_port = command_line->GetSwitchValue("remote-debugging-port");
+  if (!debugger_port.empty()) {
+    int port = atoi(debugger_port.ToString().c_str());
```
vickramdhawal: using atom seems acceptable in this case since the string is coming from the CommandLine handler. Better would be to use strtol.

appshell/cefclient.cpp
```diff
@@ -19,6 +19,7 @@
 #include "config.h"
 
 CefRefPtr<ClientHandler> g_handler;
+int g_remote_debugging_port = 0;
```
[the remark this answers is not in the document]
vickramdhawal: Since, we are going to use atom or strtol further, as suggested by you using int is better for comparisons.

appshell/config.h
```diff
@@ -82,7 +82,7 @@
 
 #endif
 
-#define REMOTE_DEBUGGING_PORT 9234
+extern int g_remote_debugging_port;
```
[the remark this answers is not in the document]
jha-g: Done.

appshell/cefclient.cpp
jha-g: Changed parsing to use `strtol`.

appshell/cefclient.cpp
```diff
@@ -95,7 +96,21 @@ void AppGetSettings(CefSettings& settings, CefRefPtr<CefCommandLine> command_lin
       command_line->GetSwitchValue(client::switches::kJavascriptFlags);
     
   // Enable dev tools
-  settings.remote_debugging_port = REMOTE_DEBUGGING_PORT;
+  CefString debugger_port = command_line->GetSwitchValue("remote-debugging-port");
+  if (!debugger_port.empty()) {
+    int port = atoi(debugger_port.ToString().c_str());
+    static const int max_port_num = 65535;
+    static const int max_reserved_port_num = 1024;
```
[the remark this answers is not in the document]
jha-g: Using `strtol` for integer parsing, so moved to `signed long`.

jha-g:
Addressed review comments. It is not possible to use `g_handler` to hold remote-debugging-port as `g_handler` does not get initialized while we are parsing command line.

appshell/cefclient.cpp
```diff
@@ -95,7 +97,25 @@ void AppGetSettings(CefSettings& settings, CefRefPtr<CefCommandLine> command_lin
...
   // Enable dev tools
-  settings.remote_debugging_port = REMOTE_DEBUGGING_PORT;
+  CefString debugger_port = command_line->GetSwitchValue("remote-debugging-port");
+  if (!debugger_port.empty()) {
+    const long port = strtol(debugger_port.ToString().c_str(), NULL, 10);
+    if (errno == ERANGE) {
+        LOG(ERROR) << "Error while parsing remote-debugging-port arg: "<< debugger_port.ToString();
+        errno = 0;
+    }
+    static const long max_port_num = 65535;
```
vickramdhawal: all of this should then be in the else case if errno is not ERANGE, since we know the port was not parsed properly and comparisons further may not give desired results.

jha-g: Done.

PR #681, opened 2020-03-30
Wrong UserAgent for windows 10 as Compatibility Manifest were not getting applied in windows build

niteskum:
Brackets windows version compatibility configuration are added in compatibility.manifest file,
but this compatibility configuration were not getting applied in build as this "compatibility.manifest" was not added in "AdditionalManifestFiles".

Added "compatibility.manifest" file in "AdditionalManifestFiles" , Now correct UserAgent string are coming for windows 10.

@narayani28 @jha-g please review.

### DIFF
--- a/appshell.gyp
+++ b/appshell.gyp
@@ -116,6 +116,7 @@
             'VCManifestTool': {
               'AdditionalManifestFiles': [
                 'appshell.exe.manifest',
+                '>(win_exe_compatibility_manifest)',
               ],
             },
           },


niteskum:
> Can you add reference links for the syntax added for the new file?

I just checked how it added in other open source project(e.g chromium) :
https://chromium.googlesource.com/chromium/chromium/+/trunk/build/common.gypi#5368

niteskum:
> Should you check for valid manifest file similar to " ['_type=="executable" and ">(win_exe_compatibility_manifest)"!=""', " as mentioned in the reference link?

in reference link they only have compatibility_manifest. but in our case we have two manifest so if "win_exe_compatibility_manifest" is empty ,we still like to continue with just one manifest (appshell.exe.manifest). having empty variable "win_exe_compatibility_manifest" will not cause any issue.

narayani28:
have you verified that if the compatibility.manifest didnt exist on the machine, does the build succeed and launch?

niteskum:
> have you verified that if the compatibility.manifest didnt exist on the machine, does the build succeed and launch?

if compatibility.manifest does not exist or does not have right permission or invalid, build will fail. I don't see there is anyway we can have check for existence and validity of a file in config .

at Line no 93 : we set "win_exe_compatibility_manifest" variable to 'compatibility.manifest'. what I verified that if I set variable 'win_exe_compatibility_manifest' to empty , build succeed and launch.

in reference link also check was if 'win_exe_compatibility_manifest' not empty

narayani28:
LGTM

niteskum:
@nethip Could you please review this PR?

PR #684, opened 2021-03-01
EOL Notification

swmitra:
EOL Notification


M3kH:
Thanks to have started what end to be the era of new browser-in-app applications.
🙏  I enjoy the evolution of the project from every angle.

